PR #18, opened 2017-03-19
use typedef structs that are compatible with both c and c++

aoloe:
the current forward declarations do not work with g++.

### DIFF
--- a/md4c/md4c.h
+++ b/md4c/md4c.h
@@ -54,8 +54,7 @@ typedef unsigned MD_OFFSET;
 
 /* Block represents a part of document hierarchy structure like a paragraph
  * or list item. */
-typedef enum MD_BLOCKTYPE_tag MD_BLOCKTYPE;
-enum MD_BLOCKTYPE_tag {
+typedef enum MD_BLOCKTYPE {
     /* <body>...</body> */
     MD_BLOCK_DOC = 0,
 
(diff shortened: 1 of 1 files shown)


mity:
Thanks. 

The diff itself is ok.

But I noticed the commit belongs to author `a.l.e` which github does not recognize as belonging to your account. It's not a problem for me, but maybe you might like to fix it before I merge it? If it's not important for you, just let me know and I'll merge as it is.

aoloe:
well, the link was not that important... but i took it as a challenge and a/ fixed the gitconfig on this computer b/ discovered how to change the name attached to the last commit!

anyway: thanks for your work on md4c!
i could get a very basic c++ sample that correctly binds the callbacks.
during the next few days, i'll try to make it more c++-alike and -- if you welcome it -- i'll share the samples so that you eventually can use them for the documentation.

mity:
;-) Again the e-mail is not recognized. 

If you want to fix it, you have either to set to your git config e-mail address already registered with your github account, or alternatively you may add more e-mails in your github account settings. See <https://github.com/settings/emails>.

aoloe:
oufh... so much work on a sunday night...

i've now added the other address to github : - )
thanks for the instructions!

mity:
Thanks.

Issue #34, opened 2018-04-11
Lots of warnings


markand:
Hello,

Thank you for this awesome library, I just get a lots of warnings when compiling with `-Wall -Wextra`:

```
$ gcc -std=c99 -Wall -Wextra md4c.c -c
md4c.c: In function ‘md_decode_unicode’:
md4c.c:844:52: warning: unused parameter ‘str_size’ [-Wunused-parameter]
     md_decode_unicode(const CHAR* str, OFF off, SZ str_size, SZ* p_size)
                                                    ^~~~~~~~
md4c.c: In function ‘md_merge_lines’:
md4c.c:864:73: warning: unused parameter ‘n_lines’ [-Wunused-parameter]
 md_merge_lines(MD_CTX* ctx, OFF beg, OFF end, const MD_LINE* lines, int n_lines,
                                                                         ^~~~~~~
md4c.c: In function ‘md_is_hex_entity_contents’:
md4c.c:1275:35: warning: unused parameter ‘ctx’ [-Wunused-parameter]
 md_is_hex_entity_contents(MD_CTX* ctx, const CHAR* text, OFF beg, OFF max_end, OFF* p_end)
                                   ^~~
md4c.c: In function ‘md_is_dec_entity_contents’:
md4c.c:1291:35: warning: unused parameter ‘ctx’ [-Wunused-parameter]
 md_is_dec_entity_contents(MD_CTX* ctx, const CHAR* text, OFF beg, OFF max_end, OFF* p_end)
                                   ^~~
md4c.c: In function ‘md_is_named_entity_contents’:
md4c.c:1307:37: warning: unused parameter ‘ctx’ [-Wunused-parameter]
 md_is_named_entity_contents(MD_CTX* ctx, const CHAR* text, OFF beg, OFF max_end, OFF* p_end)
                                     ^~~
md4c.c: In function ‘md_free_attribute’:
md4c.c:1412:27: warning: unused parameter ‘ctx’ [-Wunused-parameter]
 md_free_attribute(MD_CTX* ctx, MD_ATTRIBUTE_BUILD* build)
                           ^~~
md4c.c: In function ‘md_rollback’:
md4c.c:2618:18: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for(i = 1; i < SIZEOF_ARRAY(ctx->mark_chains); i++) {
                  ^
md4c.c: In function ‘md_build_mark_char_map’:
[...]

PR #35, opened 2018-04-11
Fix/prevent gcc 7.2 extra warnings

mity:
Addresses #34


vtorri:
declare i as size_t instead of plenty of casts in 7d9cd9a

mity:
Superseded by better #141

PR #46, opened 2019-02-02
prevent buffer overflow in md_is_entity_str

ec1oud:
Found by fuzzing: https://bugreports.qt.io/browse/QTBUG-72937
CH(off) reads past the end of ctx->text if off >= ctx->size


mity:
Thanks for it.

But I believe the provided patch is not correct: The argument `text` in `md_is_entity_str()` can point anywhere into the document so comparing the offset in this string with `ctx->size` is not in general the right thing to do.

I successfully reproduced the issue and opened #47 for it. Working on it.

PR #50, opened 2019-02-05
Implement task lists.

### LINKED ISSUE #30: Feature request: parse GitHub checkbox syntax
Another reason I want to add Markdown support to Qt is to enable combined notes and TODO lists like Evernote has, to be able to edit GitHub README.md files, etc.  

https://github.com/blog/1825-task-lists-in-all-markdown-documents

As they show there, the syntax is

```
### Solar System Exploration, 1950s – 1960s

- [ ] Mercury
- [x] Venus
- [x] Earth (Orbit/Moon)
- [x] Mars
- [ ] Jupiter
- [ ] Saturn
- [ ] Uranus
- [ ] Neptune
- [ ] Comet Haley
```

### Solar System Exploration, 1950s – 1960s

- [ ] Mercury
- [x] Venus
- [x] Earth (Orbit/Moon)
- [x] Mars
- [ ] Jupiter
- [ ] Saturn
- [ ] Uranus
- [ ] Neptune
- [ ] Comet Haley

It won't be quite so easy to add support for checking and unchecking them to any Qt text-editing component, but I'd like to try.

Currently for my personal todo lists and shopping lists etc., I use todo.txt format (https://f-droid.org/packages/nl.mpcjanssen.simpletask/ and https://f-droid.org/packages/com.nutomic.syncthingandroid/ are a good combination for this); but it has the disadvantage of not being able to mix free-form notes with the checklists.  So I think markdown will be better.

mity:
Implementation of GitHub-style task lists.

Fixes #30.

To do:
 * [x] More testing (passes tests in `tests/tasklists.txt` and tested with `afl-fuzz` for ~12 hours)
 * [x] Clarify behavior for `* [x] * [ ] * [x] foo` (see https://github.com/github/cmark-gfm/issues/134)
 * [ ] Wait for ACK by @ec1oud, given all the ongoing discussion below

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 New features:
 
+ * Add extension for GitHub-style task lists:
+
+   ```
+    * [x] foo
+    * [x] bar
+    * [ ] baz
+   ```
+
  * Renamed structure `MD_RENDERER` to `MD_PARSER` and refactorize its contents
    a little bit. Note this is source-level incompatible and initialization code
    in apps may need to be updated.
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ extensions and/or deviations from the specification.
 
  * With the flag `MD_FLAG_TABLES`, GitHub-style tables are supported.
 
+ * With the flag `MD_FLAG_TASKLISTS`, GitHub-style task lists are supported.
+
  * With the flag `MD_FLAG_STRIKETHROUGH`, strike-through spans are enabled
    (text enclosed in tilde marks, e.g. `~foo bar~`).
 
--- a/md2html/md2html.c
+++ b/md2html/md2html.c
@@ -208,6 +208,7 @@ static const option cmdline_options[] = {
     { "fcollapse-whitespace",        0,  'W', OPTION_ARG_NONE },
     { "ftables",                     0,  'T', OPTION_ARG_NONE },
     { "fstrikethrough",              0,  'S', OPTION_ARG_NONE },
+    { "ftasklists",                  0,  'X', OPTION_ARG_NONE },
     { 0 }
 };
 
@@ -255,6 +256,7 @@ usage(void)
         "      --fno-html       Same as --fno-html-blocks --fno-html-spans\n"
         "      --ftables        Enable tables\n"
         "      --fstrikethrough Enable strikethrough spans\n"
+        "      --ftasklists     Enable task lists\n"
     );
 }
 
@@ -302,6 +304,7 @@ cmdline_callback(int opt, char const* value, void* data)
         case 'V':   parser_flags |= MD_FLAG_PERMISSIVEAUTOLINKS; break;
         case 'T':   parser_flags |= MD_FLAG_TABLES; break;
         case 'S':   parser_flags |= MD_FLAG_STRIKETHROUGH; break;
+        case 'X':   parser_flags |= MD_FLAG_TASKLISTS; break;
 
         default:
             fprintf(stderr, "Illegal option: %s\n", value);
--- a/md2html/render_html.c
+++ b/md2html/render_html.c
@@ -268,6 +268,20 @@ render_open_ol_block(MD_RENDER_HTML* r, const MD_BLOCK_OL_DETAIL* det)
 }
 
 static void
+render_open_li_block(MD_RENDER_HTML* r, const MD_BLOCK_LI_DETAIL* det)
+{
+    if(det->is_task) {
+        RENDER_LITERAL(r, "<li class=\"task-list-item\">"
+                          "<input type=\"checkbox\" class=\"task-list-item-checkbox\" disabled");
+        if(det->task_mark == 'x' || det->task_mark == 'X')
+            RENDER_LITERAL(r, " checked");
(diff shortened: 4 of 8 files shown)


ec1oud:
I tried but it's not working yet... I must be missing something.  `is_task` is false and `mark` is empty.

https://codereview.qt-project.org/#/c/252383/

Why is it that MD_BLOCK_UL_DETAIL has a `mark` and MD_BLOCK_LI_DETAIL has `task_mark`?  Wouldn't it be more flexible to have `mark` in MD_BLOCK_LI_DETAIL only, in case someone mixes marks within the same list?  Or is that not allowed?

mity:
> I tried but it's not working yet... I must be missing something. `is_task` is false and `mark` is empty.

It already passes all the tests in the `tests/tasklists.txt` so `md2html` gets the detailed data correctly. Did you use `MD_FLAG_TASKLISTS`? The feature is an extension beyond the standard CommonMark specification.

> Why is it that MD_BLOCK_UL_DETAIL has a `mark` and MD_BLOCK_LI_DETAIL has `task_mark`? Wouldn't it be more flexible to have `mark` in MD_BLOCK_LI_DETAIL only, in case someone mixes marks within the same list? Or is that not allowed?

Unordered list must use have the same mark (for example `*` or `-`) for all the items or it is not considered one list. Similarly, ordered lists must use the same char after the number (`.` or `)`) for all items. So, it really is property of complete list, not of every list item. 

Also this was part of the API before working on this feature.

Duplicating the list detail data in the item detail is potentially possible, but we would have to distinguish ordered list item and unordered list items, and I am not sure whether it is worth of the complication.
(Yes, if you haven't notice, ordered list can also have task items. I follow cmark-gfm here.)

The task marks are different in that it is really property of every item. The task items and ordinary items may be mixed in a single list.

ec1oud:
Yeah I forgot that the feature flags are in more than one place in my patches... I'd better try to come up with a way to define them in only one place.  So it works after fixing that.

Do you think that MD_FLAG_TASKLISTS and such will be there for the long term?  I could remove those from the Qt API and just offer the two dialects, and allow the meaning to change over time.  But it will mean Qt users can't disable features within one dialect.  OTOH keeping all those feature flags means when you add one, Qt should add it too; and if you ever remove one, we have to deprecate it in Qt API (because it will no longer do anything), but we can't remove it.

Also, most are positive (MD_FLAG_PERMISSIVEWWWAUTOLINKS) and some are negative (MD_FLAG_NOINDENTEDCODEBLOCKS).  Are you sure you want to keep them that way?

mity:
> I could remove those from the Qt API and just offer the two dialects, and allow the meaning to change over time.

IDK what makes best sense on Qt API level. 

But consider this scenario:
  1. GitHub tomorrow adds new syntax feature called XY.
  2. MD4C eventually implements it, adds new `MD_FLAG_XY` and changes `MD_DIALECT_GITHUB` to include it.
  3. New rebuilds of Qt (if Qt uses the macro `MD_DIALECT_GITHUB`) could now get new block or span types, it is not ready for.

IMHO the safest way from compatibility point of view is, if Qt calls MD4C with the particular sets of flags directly, no matter what Qt offers to apps in its API.

> Also, most are positive (MD_FLAG_PERMISSIVEWWWAUTOLINKS) and some are negative (MD_FLAG_NOINDENTEDCODEBLOCKS). Are you sure you want to keep them that way?

The rule so far is `flags == 0` means CommonMark specification. That core is safest to use.

The negatives are for things some people see as misfeatures in the specification, and there is some demand/reasoning in order to allow suppressing it. Take `MD_FLAG_NOHTML` as an example. Markdown syntax originally began as a more comfortable way how to write HTML. But if you use it in any other context, the raw HTML blocks/inlines often do not make much sense. 

Similarly `MD_FLAG_NOINDENTEDCODEBLOCKS` is seen by some people as bad idea given the complicated indentation rules in e.g. nested lists; and there is fine alternative in the fenced code blocks.

The extensions are more likely to change to bring them to the original begavior (e.g. github markdown) if some differences are observed. The situation is that the extensions typically have no or very poor specification; or that there are bigger differences among various Markdown implementations. There is sort of Babel situation, unfortunately.

> Do you think that MD_FLAG_TASKLISTS and such will be there for the long term?

In short, yes, I think the flags will stick for the long time.

I hope, that at least the most popular extensions (tables, maybe task lists, permissive auto-links) may likely end up in the CommonMark specification. But given the slow progress on it so far, it may take years. (No such change happened since I started the work on MD4C about 3 years ago. In this period there were only changes how to interpret some corner cases and clarifications in spec wording.)

If/when that happens, it shall likely deserve bump of the major version; or to require `abi_version=1` to identify that the caller is ready to get broader sets of block and/or span types into the callbacks. In either case we can keep the flag macro there (it would just become noop).

mity:
@ec1oud Are you satisfied with my previous answers? As this thread affects the abi maintenance policy, I would like to be sure we have reached a consensus on this before making any release with the shared lib support.

Also, if you are ok with the interface for the task lists, I'd merge it before the release. If not, it can be left for some future release.

ec1oud:
I think it's all good, I'm just trying to get checklists actually working.  Getting close though.

Issue #61, opened 2019-03-10
Flag MD_FLAG_NOHTMLSPANS disables also autolinks


mity:
`MD_FLAG_NOHTMLSPANS` is supposed to disable inline raw HTML but nothing else. It disables also (standard CommonMark) autolinks:

``` 
$ echo '<http://google.com>' | md2html 
<p><a href="http://google.com">http://google.com</a></p>

$ echo '<http://google.com>' | md2html --fno-html-spans
<p>&lt;http://google.com&gt;</p>
```

Issue #59, opened 2019-03-10
Parsing '\``\``\``\``...' takes quadratic time


mity:
```
$ python -c 'print("\\``" * 20000)' | time md2html > /dev/null
0.81user 0.00system 0:00.85elapsed 95%CPU (0avgtext+0avgdata 2480maxresident)k
0inputs+0outputs (0major+333minor)pagefaults 0swaps
$ python -c 'print("\\``" * 40000)' | time md2html > /dev/null
3.70user 0.00system 0:03.76elapsed 98%CPU (0avgtext+0avgdata 3240maxresident)k
0inputs+0outputs (0major+551minor)pagefaults 0swaps
$ python -c 'print("\\``" * 80000)' | time md2html > /dev/null
15.23user 0.00system 0:15.35elapsed 99%CPU (0avgtext+0avgdata 5008maxresident)k
0inputs+0outputs (0major+991minor)pagefaults 0swaps
```

(the 2nd case from #57)

Issue #58, opened 2019-03-10
Parsing '<><><><>...' takes quadratic time


mity:
```
$ python -c 'print("<>" * 10000)' | time md2html > /dev/null
1.41user 0.00system 0:01.45elapsed 96%CPU (0avgtext+0avgdata 1968maxresident)k
0inputs+0outputs (0major+229minor)pagefaults 0swaps
$ python -c 'print("<>" * 20000)' | time md2html > /dev/null
4.80user 0.00system 0:04.84elapsed 99%CPU (0avgtext+0avgdata 2608maxresident)k
0inputs+0outputs (0major+365minor)pagefaults 0swaps
$ python -c 'print("<>" * 40000)' | time md2html > /dev/null
19.52user 0.00system 0:19.65elapsed 99%CPU (0avgtext+0avgdata 3528maxresident)k
0inputs+0outputs (0major+640minor)pagefaults 0swaps
```

(the 1st case from #57)

PR #62, opened 2019-03-10
[WIP] Fixes

mity:
Work on fixing few issues:
 * [x] #58: Parsing '<><><><>...' takes quadratic time
 * [x] #59: Parsing '\``\``\``\``...' takes quadratic time
 * [x] #61: Flag `MD_FLAG_NOHTMLSPANS` disables also autolinks

### DIFF
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -123,19 +123,20 @@ struct MD_CTX_tag {
 #endif
 
     /* For resolving of inline spans. */
-    MD_MARKCHAIN mark_chains[8];
+    MD_MARKCHAIN mark_chains[7];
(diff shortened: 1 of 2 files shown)


mity:
The commit 1599809 caused regression in [Example 322](https://spec.commonmark.org/0.28/#example-322). This is a calculated trade-off which should be resolved in the upcoming work.

mity:
The commit 9107ffc caused regression in Example 325 and Example 572. This is a calculated trade-off which should be resolved in the upcoming work: We need to move autolink also into `md_collect_marks()`.

mity:
Aborted. Not done the right way.

PR #67, opened 2019-03-26
[WIP] Updates for spec 0.29

mity:
Given the recent activity in https://github.com/commonmark/CommonMark, I guess specification 0.29 is quite close. Lets be ready for it.

(Won't be merged into master until the 0.29 specification is finally released.)

### DIFF
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -2760,21 +2760,27 @@ md_build_mark_char_map(MD_CTX* ctx)
 #define CODESPAN_MARK_MAXLEN    32
 
 static int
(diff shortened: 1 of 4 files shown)


mity:
The commit 6e49193 still may need some attention, as we still disagree with the example in the specs, just in a different way. See https://github.com/commonmark/cmark/issues/193#issuecomment-476589873

mity:
As of https://github.com/commonmark/CommonMark/commit/c1e0183685e22282cbf5651970200fe56a712002, this PR should be complete and we pass all the specs tests.

PR #75, opened 2019-05-05
Build: Fix a few sign-comparison issues

aplgithub:
This will fix a couple of issues where the GCC compiler (in strict
mode) is complaining with "error: comparison between signed and
unsigned integer expressions [-Werror=sign-compare]".


mity:
LGTM. Thanks.

PR #85, opened 2019-06-24
Add man page for md2html program

craigbarnes:


mity:
That's awesome. Thank you.

Could you please also add something into `md2html/CMakeLists.txt` to enable installing it? Would be better to merge it together as one changeset.

I suppose something like 

``` CMake
install(FILES "md2html.1" DESTINATION "${CMAKE_INSTALL_MANDIR}/man1")
```

should work.

craigbarnes:
Updated `md2html/CMakeLists.txt` with the exact command you suggested. I've tested it with `make install DESTDIR=./tmp` and it seems to work as intended.

mity:
Merged. Thank you once more again.

Issue #86, opened 2019-07-01
Support LaTeX style equations


dyedgreen:
I have a use-case where I need to support LaTeX style equations (e.g. `$a^2+b^2=c^2$`, `$$\int_a^b x dx$$` for display).

I managed to cobble something together to have an extension that emits these as spans. The extension can be activated using a flag, much like the ~strike through~ extension. I haven't written any tests or added support for it in the included html renderer (in part because I need a custom renderer and in part because there is no standard way of displaying LaTeX equations in html).

If you want, I can create a pull requests and with a few pointers I could add in the support in the renderer and add some tests.

mity:
I understand there is quite a large demand within Markdown user community for the feature. E.g. https://talk.commonmark.org/search?q=math shows that clearly. So yes, supporting it would be great thing.

That said, I have several points though:

1. It seems there is no real consensus about the syntax yet (if ever there will be), see e.g. https://github.com/cben/mathdown/wiki/math-in-markdown. 
   
   Given the large need, I am still fine to add support for it, but it would be good if we add ideally a syntax which seems to be the most common one. Do you know current state of this support in other Markdown implementations?

2. Also, even though you mentioned just spans, there is clearly also demand for math blocks as well. So the syntax should be somehow friendly for eventual expansion in that direction. I.e. the math syntax inside the block and the span should ideally be the same once the blocks are supported as well, so that only the outer decoration would differ. Is it fine from this POV?

3. What about false positives? Can the extension deal reasonably with `$` when it isn't meant to mark LaTeX equation? Consider e.g. this:
   ```
   I thought the ticket was $20, but instead it was $25.
   ```

   [Pandoc uses some heuristics for this](https://talk.commonmark.org/t/mathjax-extension-for-latex-equations/698/7) so we can likely follow it?

4. Your 2nd example includes backslash. Backslashes have special meaning in Markdown so we have to be careful so the two do not interact badly.

5. Any new syntax feature **must** have its counterpart in the HTML renderer because all our testing is based on it. However it would be completely fine if it generates something super-simple as e.g. `<span class="math inline">\(1+1=2\)</span>`, [similarly as Pandoc does](https://pandoc.org/try/?text=I+thought+the+ticket+was+%2420%2C+but+instead+it+was+%2425.%0A%0AThis+equation+%241%2B1%3D2%24+is+rendered+as+math.&from=markdown&to=html).

(Sorry for possibly banal questions. I am not LaTeX math user myself.)

[...]

Issue #71, opened 2019-04-16
compiler for INTEGRITY OS complains about non-constant initialization of MD_LINE


ec1oud:
From https://testresults.qt.io/logs/qt/qtbase/7e01267e18c12cf19d8cf7487126efeeaeb84b23/LinuxRHEL_7_4x86_64INTEGRITYINTEGRITY_11_04armv7GCCqtci-linux-RHEL-7.4-x86_64-fa00a5Release_DisableTests_OpenGLES2_NoUseGoldLinker/2a48a3b7dcb9652b1532a3ad14de7580b5b9b1ee/build_1555421647/log.txt.gz

```
"../3rdparty/md4c/md4c.c", line 3378: warning #68-D: integer conversion
           resulted in a change of sign
                   OFF inline_link_end = -1;
                                         ^
 
 "../3rdparty/md4c/md4c.c", line 4155: error #28: expression must have a
           constant value
       MD_LINE line = { beg, end };
                        ^
 
 "../3rdparty/md4c/md4c.c", line 4155: error #28: expression must have a
           constant value
       MD_LINE line = { beg, end };
                             ^
 
 "../3rdparty/md4c/md4c.c", line 4246: error #28: expression must have a
           constant value
       MD_LINE line = { beg, beg };
                        ^
 
 "../3rdparty/md4c/md4c.c", line 4246: error #28: expression must have a
           constant value
       MD_LINE line = { beg, beg };
                             ^
 
 "../3rdparty/md4c/md4c.c", line 5189: warning #940-D: missing return statement
           at end of non-void function "md_is_html_block_end_condition"
   }
   ^
```
Host: os: Linux, osVersion: RHEL_7_4, compiler: GCC, arch: X86_64
Target: os: INTEGRITY, osVersion: INTEGRITY_11_04, compiler: GCC, arch: ARMv7

Maybe making those beg and end arguments const will satisfy it?  not sure... and I don't have a development environment for testing, so we'll have to keep trying with the CI system I guess.

mity:
And do you know what's the gcc version? (I guess these errors are from the compiler's front-end so using gcc of the same version on another platform could generate the same error output.)

ec1oud:
[0001-fixup.patch.gz](https://github.com/mity/md4c/files/3086759/0001-fixup.patch.gz)

WAG patch... I'll try with CI

ec1oud:
[...]

PR #87, opened 2019-07-02
Added support for LaTeX equations

dyedgreen:
Supports parsing for $ and $$. This
addresses #86.

### DIFF
--- a/md2html/md2html.c
+++ b/md2html/md2html.c
@@ -208,7 +208,7 @@ static const option cmdline_options[] = {
     { "fcollapse-whitespace",        0,  'W', OPTION_ARG_NONE },
     { "ftables",                     0,  'T', OPTION_ARG_NONE },
     { "fstrikethrough",              0,  'S', OPTION_ARG_NONE },
-    { "flatex",                      0,  'L', OPTION_ARG_NONE },
+    { "flatex-math",                 0,  'L', OPTION_ARG_NONE },
     { "ftasklists",                  0,  'X', OPTION_ARG_NONE },
     { 0 }
 };
@@ -257,7 +257,7 @@ usage(void)
         "      --fno-html       Same as --fno-html-blocks --fno-html-spans\n"
         "      --ftables        Enable tables\n"
         "      --fstrikethrough Enable strikethrough spans\n"
-        "      --flatex         Enable LaTeX equation spans\n"
+        "      --flatex-math    Enable LaTeX style mathematics spans (e.g. $a+b=c$ becomes <equation>a+b=c</equation>)\n"
         "      --ftasklists     Enable task lists\n"
     );
 }
@@ -306,7 +306,7 @@ cmdline_callback(int opt, char const* value, void* data)
         case 'V':   parser_flags |= MD_FLAG_PERMISSIVEAUTOLINKS; break;
         case 'T':   parser_flags |= MD_FLAG_TABLES; break;
         case 'S':   parser_flags |= MD_FLAG_STRIKETHROUGH; break;
-        case 'L':   parser_flags |= MD_FLAG_LATEX; break;
+        case 'L':   parser_flags |= MD_FLAG_LATEXMATHSPANS; break;
         case 'X':   parser_flags |= MD_FLAG_TASKLISTS; break;
 
         default:
--- a/md2html/md2html.c
+++ b/md2html/md2html.c
@@ -208,6 +208,7 @@ static const option cmdline_options[] = {
     { "fcollapse-whitespace",        0,  'W', OPTION_ARG_NONE },
     { "ftables",                     0,  'T', OPTION_ARG_NONE },
     { "fstrikethrough",              0,  'S', OPTION_ARG_NONE },
+    { "flatex",                      0,  'L', OPTION_ARG_NONE },
     { "ftasklists",                  0,  'X', OPTION_ARG_NONE },
     { 0 }
 };
@@ -256,6 +257,7 @@ usage(void)
         "      --fno-html       Same as --fno-html-blocks --fno-html-spans\n"
         "      --ftables        Enable tables\n"
         "      --fstrikethrough Enable strikethrough spans\n"
+        "      --flatex         Enable LaTeX equation spans\n"
         "      --ftasklists     Enable task lists\n"
     );
 }
@@ -304,6 +306,7 @@ cmdline_callback(int opt, char const* value, void* data)
         case 'V':   parser_flags |= MD_FLAG_PERMISSIVEAUTOLINKS; break;
         case 'T':   parser_flags |= MD_FLAG_TABLES; break;
         case 'S':   parser_flags |= MD_FLAG_STRIKETHROUGH; break;
+        case 'L':   parser_flags |= MD_FLAG_LATEX; break;
         case 'X':   parser_flags |= MD_FLAG_TASKLISTS; break;
 
         default:
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -141,7 +141,7 @@ struct MD_CTX_tag {
 #define BRACKET_OPENERS                         ctx->mark_chains[10]
 #define DOLLAR_OPENERS                          ctx->mark_chains[11]
 #define OPENERS_CHAIN_FIRST                     2
-#define OPENERS_CHAIN_LAST                      10
+#define OPENERS_CHAIN_LAST                      11
 
     int n_table_cell_boundaries;
 
@@ -3654,17 +3654,27 @@ md_analyze_dollar(MD_CTX* ctx, int mark_index)
 {
     /* This should mimic the way inline equations work in LaTeX, so there
      * can only ever be one item in the chain (i.e. the dollars can't be
-     * nested). This is basically the same as the md_analyze_tilde function.
+     * nested). This is basically the same as the md_analyze_tilde function,
+     * except that we require matching openers and closers to be of the same
+     * length.
      *
-     * Note: Doing things this way means we are very lenient and accept and
-     * number (1,2) of $ to close an equation. */
+     * E.g.: $abc$$def$$ => abc (display equation) def (end equation) */
     if(DOLLAR_OPENERS.head >= 0) {
-        /* The chain already contains an opener, so we are the closer. */
-        int opener_index = DOLLAR_OPENERS.head;
+        /* If the potential closer has a non-matching number of $, discard */
+        MD_MARK open = ctx->marks[DOLLAR_OPENERS.head];
+        MD_MARK close = ctx->marks[mark_index];
 
+        int opener_index = DOLLAR_OPENERS.head;
         md_rollback(ctx, opener_index, mark_index, MD_ROLLBACK_CROSSING);
-        md_resolve_range(ctx, &DOLLAR_OPENERS, opener_index, mark_index);
+        if (open.end - open.beg == close.end - close.beg) {
+            /* We are the matching closer */
+            md_resolve_range(ctx, &DOLLAR_OPENERS, opener_index, mark_index);
+        } else {
+            /* We don't match the opener, so discard old opener and insert as opener */
+            md_mark_chain_append(ctx, &DOLLAR_OPENERS, mark_index);
+        }
     } else {
+        /* No unmatched openers, so we are opener */
         md_mark_chain_append(ctx, &DOLLAR_OPENERS, mark_index);
     }
 }
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -3661,12 +3661,12 @@ md_analyze_dollar(MD_CTX* ctx, int mark_index)
      * E.g.: $abc$$def$$ => abc (display equation) def (end equation) */
     if(DOLLAR_OPENERS.head >= 0) {
         /* If the potential closer has a non-matching number of $, discard */
-        MD_MARK open = ctx->marks[DOLLAR_OPENERS.head];
-        MD_MARK close = ctx->marks[mark_index];
+        MD_MARK* open = &ctx->marks[DOLLAR_OPENERS.head];
+        MD_MARK* close = &ctx->marks[mark_index];
 
         int opener_index = DOLLAR_OPENERS.head;
         md_rollback(ctx, opener_index, mark_index, MD_ROLLBACK_CROSSING);
-        if (open.end - open.beg == close.end - close.beg) {
+        if (open->end - open->beg == close->end - close->beg) {
             /* We are the matching closer */
             md_resolve_range(ctx, &DOLLAR_OPENERS, opener_index, mark_index);
         } else {
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -4029,7 +4029,7 @@ md_process_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
                 case '$':
                     if(mark->flags & MD_MARK_OPENER) {
                         MD_ENTER_SPAN((mark->end - off) % 2 ? MD_SPAN_LATEXMATH : MD_SPAN_LATEXMATH_DISPLAY, NULL);
-                        text_type = MD_TEXT_CODE; /* LaTeX should be read as code */
+                        text_type = MD_TEXT_LATEXMATH;
                     } else {
                         MD_LEAVE_SPAN((mark->end - off) % 2 ? MD_SPAN_LATEXMATH : MD_SPAN_LATEXMATH_DISPLAY, NULL);
                         text_type = MD_TEXT_NORMAL;
@@ -4134,12 +4134,17 @@ md_process_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
             if(off >= end)
                 break;
 
-            if(text_type == MD_TEXT_CODE) {
+            if(text_type == MD_TEXT_CODE || text_type == MD_TEXT_LATEXMATH) {
                 OFF tmp;
 
                 MD_ASSERT(prev_mark != NULL);
-                MD_ASSERT(prev_mark->ch == '`'  &&  (prev_mark->flags & MD_MARK_OPENER));
-                MD_ASSERT(mark->ch == '`'  &&  (mark->flags & MD_MARK_CLOSER));
+                if (text_type == MD_TEXT_CODE) {
+                    MD_ASSERT(prev_mark->ch == '`'  &&  (prev_mark->flags & MD_MARK_OPENER));
+                    MD_ASSERT(mark->ch == '`'  &&  (mark->flags & MD_MARK_CLOSER));
+                } else if (text_type == MD_TEXT_LATEXMATH) {
+                    MD_ASSERT(prev_mark->ch == '$'  &&  (prev_mark->flags & MD_MARK_OPENER));
+                    MD_ASSERT(mark->ch == '$'  &&  (mark->flags & MD_MARK_CLOSER));
+                }
 
                 /* Inside a code span, trailing line whitespace has to be
                  * outputted. */
@@ -4147,11 +4152,11 @@ md_process_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
                 while(off < ctx->size  &&  ISBLANK(off))
                     off++;
                 if(off > tmp)
-                    MD_TEXT(MD_TEXT_CODE, STR(tmp), off-tmp);
+                    MD_TEXT(text_type, STR(tmp), off-tmp);
 
                 /* and new lines are transformed into single spaces. */
                 if(prev_mark->end < off  &&  off < mark->beg)
-                    MD_TEXT(MD_TEXT_CODE, _T(" "), 1);
+                    MD_TEXT(text_type, _T(" "), 1);
 
             } else if(text_type == MD_TEXT_HTML) {
                 /* Inside raw HTML, we output the new line verbatim, including
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -3665,7 +3665,7 @@ md_analyze_dollar(MD_CTX* ctx, int mark_index)
         MD_MARK* close = &ctx->marks[mark_index];
 
         int opener_index = DOLLAR_OPENERS.head;
-        md_rollback(ctx, opener_index, mark_index, MD_ROLLBACK_CROSSING);
+        md_rollback(ctx, opener_index, mark_index, MD_ROLLBACK_ALL);
         if (open->end - open->beg == close->end - close->beg) {
             /* We are the matching closer */
             md_resolve_range(ctx, &DOLLAR_OPENERS, opener_index, mark_index);
--- a/md4c/md4c.c
+++ b/md4c/md4c.c
@@ -127,7 +127,7 @@ struct MD_CTX_tag {
 #endif
 
     /* For resolving of inline spans. */
-    MD_MARKCHAIN mark_chains[11];
+    MD_MARKCHAIN mark_chains[12];
 #define PTR_CHAIN                               ctx->mark_chains[0]
 #define TABLECELLBOUNDARIES                     ctx->mark_chains[1]
 #define ASTERISK_OPENERS_extraword_mod3_0       ctx->mark_chains[2]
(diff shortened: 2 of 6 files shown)


md4c/md4c.c
```diff
@@ -3251,6 +3255,21 @@ md_collect_marks(MD_CTX* ctx, const MD_LINE* lines, int n_lines, int table_mode)
                 continue;
             }
 
+            /* A potential equation start/end */
+            if(ch == _T('$')) {
+                // We can have at most two consecutive $ signs,
+                // where two dollar signs signify a display equation
```
mity: Please use the old C style comment `/* ... */`.

md4c/md4c.c
```diff
@@ -3974,6 +4016,16 @@ md_process_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
                         MD_LEAVE_SPAN(MD_SPAN_DEL, NULL);
                     break;
 
+                case '$':
+                    if(mark->flags & MD_MARK_OPENER) {
+                        MD_ENTER_SPAN((mark->end - off) % 2 ? MD_SPAN_LATEX : MD_SPAN_LATEX_DISPLAY, NULL);
+                        text_type = MD_TEXT_CODE; // LaTeX should be read as code
```
mity: Please use `/* ... */` comment style.

md4c/md4c.h
```diff
@@ -129,7 +129,13 @@ typedef enum MD_SPANTYPE {
...
      * Note: Recognized only when MD_FLAG_STRIKETHROUGH is enabled.
      */
-    MD_SPAN_DEL
+    MD_SPAN_DEL,
+
+    /* For recognizing inline ($) and display ($$) equations
+     * Note: Recognized only when MD_FLAG_LATEX is enabled.
+     */
+    MD_SPAN_LATEX,
+    MD_SPAN_LATEX_DISPLAY
```
mity: Please rename to `MD_SPAN_LATEXMATH` and `MD_SPAN_LATEXMATH_DISPLAY`. As far as I can understand it, this does not allow embedding any LaTeX snippets, only math ones, right?

mity:
Generally, the approach seems in the right direction to me.

md2html/md2html.c
```diff
@@ -208,6 +208,7 @@ static const option cmdline_options[] = {
     { "fcollapse-whitespace",        0,  'W', OPTION_ARG_NONE },
     { "ftables",                     0,  'T', OPTION_ARG_NONE },
     { "fstrikethrough",              0,  'S', OPTION_ARG_NONE },
+    { "flatex",                      0,  'L', OPTION_ARG_NONE },
```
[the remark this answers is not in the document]
dyedgreen: Makes sense

md2html/md2html.c
```diff
@@ -256,6 +257,7 @@ usage(void)
         "      --fno-html       Same as --fno-html-blocks --fno-html-spans\n"
         "      --ftables        Enable tables\n"
         "      --fstrikethrough Enable strikethrough spans\n"
+        "      --flatex         Enable LaTeX equation spans\n"
```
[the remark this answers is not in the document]
dyedgreen: See the new commit below. Is that better?

md4c/md4c.h
```diff
@@ -275,6 +281,7 @@ typedef struct MD_SPAN_IMG_DETAIL {
 #define MD_FLAG_STRIKETHROUGH               0x0200  /* Enable strikethrough extension. */
 #define MD_FLAG_PERMISSIVEWWWAUTOLINKS      0x0400  /* Enable WWW autolinks (even without any scheme prefix, if they begin with 'www.') */
 #define MD_FLAG_TASKLISTS                   0x0800  /* Enable task list extension. */
+#define MD_FLAG_LATEX                       0x1000  /* Enable $ and $$ containing LaTeX equations. */
```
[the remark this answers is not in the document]
dyedgreen: Makes sense.

md4c/md4c.h
dyedgreen: Yes. This will only pick up on $, $$ delimiters so only on equations. So makes sense.

dyedgreen:
Please see committed changes. Hope that resolves the issues around the naming.

dyedgreen:
> Reviewed mainly the API so far, not the implementation itself.

Regarding the implementation: It basically recreates the way ~strikethrough~ is implemented, with the only difference being that text inside $/$$ is considered code and the delimiters are limited to be between 1 and 2 characters long.

mity:
Yep. The interface now looks good to me. 

> Regarding the implementation: It basically recreates the way strikethrough is implemented, with the only difference being that text inside $/$$ is considered code and the delimiters are limited to be between 1 and 2 characters long.

Noticed that. I'm just wondering whether the code span approach which takes marker length into account wouldn't be better. That would fix the strange mixing of `$...$$`, but maybe could cause some other unwanted side effects.

But I need fresh mind for that. It's 0:48 AM here and I am getting too sleepy...

md4c/md4c.c
```diff
@@ -139,6 +139,7 @@ struct MD_CTX_tag {
 #define UNDERSCORE_OPENERS                      ctx->mark_chains[8]
 #define TILDE_OPENERS                           ctx->mark_chains[9]
 #define BRACKET_OPENERS                         ctx->mark_chains[10]
+#define DOLLAR_OPENERS                          ctx->mark_chains[11]
```
mity: Update `OPENERS_CHAIN_LAST` to 11. This is needed to some code which needs to operate on all the opener chains in a generic way.

md4c/md4c.c
```diff
@@ -3630,6 +3649,26 @@ md_analyze_tilde(MD_CTX* ctx, int mark_index)
...
+    if(DOLLAR_OPENERS.head >= 0) {
+        /* The chain already contains an opener, so we are the closer. */
+        int opener_index = DOLLAR_OPENERS.head;
+
+        md_rollback(ctx, opener_index, mark_index, MD_ROLLBACK_CROSSING);
+        md_resolve_range(ctx, &DOLLAR_OPENERS, opener_index, mark_index);
+    } else {
+        md_mark_chain_append(ctx, &DOLLAR_OPENERS, mark_index);
+    }
+}
```
mity: Imho, it should not allow mixing `$` and `$$` (or vice versa). When that happens, I would simply discard the old opener, and use the new one as a potential opener instead.

That would mean that e.g.
```
$foo$$bar$$
```
sees the `bar` as a (display) equation and not `foo`. It's imho more intuitive behavior.

md4c/md4c.c
```diff
@@ -3974,6 +4016,16 @@ md_process_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
                         MD_LEAVE_SPAN(MD_SPAN_DEL, NULL);
                     break;
 
+                case '$':
+                    if(mark->flags & MD_MARK_OPENER) {
+                        MD_ENTER_SPAN((mark->end - off) % 2 ? MD_SPAN_LATEXMATH : MD_SPAN_LATEXMATH_DISPLAY, NULL);
+                        text_type = MD_TEXT_CODE; /* LaTeX should be read as code */
```
mity: I am wondering whether intorducing some `MD_TEXT_LATEXMATH` wouldn't be better. I can imagine some potential renderers may want to render differently from code.

Consider e.g. a HTML renderer which would pass the `MD_TEXT_LATEXMATH` text to some machinery which would use LaTeX to render the math into an image and embed that image in the HTML page.

Or a custom renderer which would (re)implement LaTeX math syntax and would support its rendering.

md4c/md4c.c
dyedgreen: Yes, I think that would make more sense. It can lead to confusing situations where $blub$$ ... lots of text $$some display eqn$$ would be parsed in a wired way, but I think that's overall better than the alternative (which can also lead to wired behaviour.)

md4c/md4c.c
dyedgreen: The main reason I went with `MD_TEXT_CODE` is because there are some things that happen when setting code that apply to LaTeX code as well (like not escaping anything until the span closes). If the custom renderer wants to treat LaTeX code differently from any other code, it can still easily do that by setting some flag when entering an equation (similar to how it would have to watch for opening images to correctly process the alt attribute contents).

I can see there being benefits to not having to handle this in the renderer, but at the same time it would probably make the md4c code more convoluted (? I haven't looked at how the MD_TEXT_CODE is implemented). Let me know what you think, I'll update the PR to include the other changes :)

dyedgreen:
Another comment: Should we change the name from `MD_SPAN_LATEXMATH` to `MD_SPAN_MATH`? I think that $/$$ to open equations, while it does come from LaTeX, is a very common convention now, so it may make sense to not have LaTeX in the name. Especially since this could very well be used to handle other math markup when given a $ / $$, not just LaTeX (since we treat it as text to be passed to the renderer verbatim anyway).

dyedgreen:
Another thing: did I add the tests correctly? Codecov does not seem to recognise the additional tests.

md4c/md4c.c
mity: The issue is that such renderer would have to test many conditions (is the span `MD_SPAN_LATEXMATH` **or** `MD_SPAN_LATEXMATH_DISPLAY`; and in the future also possibly some parent block if we introduce math blocks.

My 2nd argument is there is always just one parser (md4c.c), but there can potentially be many renderers, so it is better to place the complexity into the parser and make the renderers easier to write.

> I can see there being benefits to not having to handle this in the renderer, but at the same time it would probably make the md4c code more convoluted (? I haven't looked at how the MD_TEXT_CODE is implemented). Let me know what you think, I'll update the PR to include the other changes :)

I think that all would should be needed is to update the block at the end of `md_process_inlines()` which handles end of line. (Currently line number 4075 in upstream master.) 

Actually the question is whether this should be handled the same way or not.

And even if we stick with the current `MD_TEXT_CODE`, the assertions there are wrong right now: They verify we are indeed inside a code span denoted with `` ` `` chars so this:

```
this block $has a math span spread into
multiple lines$.
```

would likely explode on the assertion.

dyedgreen: > And even if we stick with the current `MD_TEXT_CODE`, the assertions there are wrong right > now: They verify we are indeed inside a code span denoted with `` ` `` chars [...]

That's a good point. I'll look into that, and introduce the additional type.

mity:
>  Should we change the name from `MD_SPAN_LATEXMATH` to `MD_SPAN_MATH`?

I don't know. If we go that way then the docs/comments can only explain the actual syntax inside the span is sort of a contract between authors of the documents and the particular renderer; and not that it follows some particular syntax. And also in such case, why to use "math" name at all? It could use even other semantics (e.g. chemistry) if it is out of scope of our specification.

But in some ways it looks to me as giving up and contributing to the Markdown babel: Every renderer could expect completely different syntax there and that's imho way to the hell.

So I would rather stick to the "LaTeX math" and specification would refer to LaTeX docs and what it says can be inside of it. Simple renderers would be recommended to output it verbatim, complex one could parse it as LaTeX does.

If someone (mis)uses it for a different purpose (or different math syntax) he will at least know he does somethign non-standard.

mity:
> Another thing: did I add the tests correctly? Codecov does not seem to recognise the additional tests.

It likely does not work in PRs correctly because there are two CI systems setup for the project: AppVeyor (for Windows MSVC builds) and Travis (for Linux builds). Only the Linux build runs the test suite and feeds CodeCov with the code coverity data, but unfortunately the Travis CI seems to ignore PRs and does not make build for them.

IDK how to make MSVC build to collect the coverity data in a way CodeCov would udnerstand (and even then it is likely bad idea to feed CodeCov from multiple data sources). I also IDK whether the build triggering in Travis can somehow be enabeld so that it does the Linux builkd when a PR is created/updated.

So it likely "fixes" itself when we eventually merge the PR.

md4c/md4c.c
mity: Well, the 1st step here is to know how it should behave on EOLs.

Quite certainly it shouldn't be handled as a normal text where EOLs are translated into `MD_TEXT_BR` or `MD_TEXT_SOFTBR`. But IDK whether translating the EOL into a space as the code spans do is the right think for the math. What do you think?

dyedgreen: I think translating EOL to spaces is the right thing to do. The normal use-case for inline spans is to reference variables in text, so a very usual pattern is:
```
Bla bla $a$ bla bla...
```
and so $...EOL...$ is very unusual anyway.

mity: Ok. Then updating that block should be quite easy:

```
if(text_type == MD_TEXT_CODE  ||  text_type == _the_new_text_type_)
```

and replacing `MD_TEXT_CODE` with text_type in that branch. 

As of the assertions, I would simply remove them as they would become too complicated otherwise. 

Please also add some multi-line math into the test file so the branch is triggered by the tests.

mity: >  so $...EOL...$ is very unusual anyway.

We cannot rely on that. Even if long equations may be unlikely, many Markdown authors want to word wrap long lines at e.g. 80 characters. And the math span can begin at the offset 75 in the text.

mity:
The latest commit is not enough. See https://github.com/mity/md4c/pull/87/commits/cc861265b5bc68b87b223c223d7ecefddda34a38#diff-7a78427587ec282a7531fcfa87d134bcR4155 and https://github.com/mity/md4c/pull/87/commits/cc861265b5bc68b87b223c223d7ecefddda34a38#diff-7a78427587ec282a7531fcfa87d134bcR4155

The both line now need to fire `text_type`, not the hard-coded `MD_TEXT_CODE`.

dyedgreen:
> The latest commit is not enough. See [cc86126#diff-7a78427587ec282a7531fcfa87d134bcR4155](https://github.com/mity/md4c/commit/cc861265b5bc68b87b223c223d7ecefddda34a38#diff-7a78427587ec282a7531fcfa87d134bcR4155) and [cc86126#diff-7a78427587ec282a7531fcfa87d134bcR4155](https://github.com/mity/md4c/commit/cc861265b5bc68b87b223c223d7ecefddda34a38#diff-7a78427587ec282a7531fcfa87d134bcR4155)
> 
> The both line now need to fire `text_type`, not the hard-coded `MD_TEXT_CODE`.
 
Just saw that, should be fixed now 😅

dyedgreen:
Just saw the unicode whitespace -.- sorry for that...

md4c/md4c.c
```diff
@@ -3630,6 +3649,36 @@ md_analyze_tilde(MD_CTX* ctx, int mark_index)
...
+    /* This should mimic the way inline equations work in LaTeX, so there
+     * can only ever be one item in the chain (i.e. the dollars can't be
+     * nested). This is basically the same as the md_analyze_tilde function,
+     * except that we require matching openers and closers to be of the same
+     * length.
+     *
+     * E.g.: $abc$$def$$ => abc (display equation) def (end equation) */
+    if(DOLLAR_OPENERS.head >= 0) {
+        /* If the potential closer has a non-matching number of $, discard */
+        MD_MARK open = ctx->marks[DOLLAR_OPENERS.head];
```
mity: Could you please use it via pointer instead of copying the whole structure? 
```
MD_MARK* opener = &ctx->marks[DOLLAR_OPENERS.head];
```

(Yes, I know it is small. But older C compilers could reject it.)

mity:
> Just saw the unicode whitespace -.- sorry for that...

Sorry. Not sure what you are talking about. Maybe because you force-pushed again. But then the previous version is more ore less inaccessible to me.

You can freely make new commits instead of force-pushes: I will likely "Squash and merge" the PR into a single commit as it all is not that big change anyway and it all is closely related.

dyedgreen:
> > Just saw the unicode whitespace -.- sorry for that...
> 
> Sorry. Not sure what you are talking about. Maybe because you force-pushed again. But then the previous version is more ore less inaccessible to me.
> 
> You can freely make new commits instead of force-pushes: I will likely "Squash and merge" the PR into a single commit as it all is not that big change anyway and it all is closely related.

No worries, there was an accidental unicode whitespace instead of a normal one and it didn't compile under windows.

md4c/md4c.c
dyedgreen: Sure, that's no problem, see new commit. Out of curiosity, why would an older compiler reject it,  is it not valid C99?

mity: MD4C is now used by (dev branch of) Qt, and they also need to build for some embedded platform(s). Those often may use very ancient versions of gcc or even some proprietary compilers.

See #71.

mity: And BTW, 

AFAIK MSVC also does not support fully C99. It only knows some of its features (as extensions to some older standard, if they really adhere some at all). Microsoft repeatedly says they do care just about C++.

So I tend to be quite defensive.

dyedgreen: Fair. I wasn't aware of that. (I rarely use anything Microsoft)

mity:
Looks good to me now.

I'll run some fuzz testing against it during the night, and it it doesn't hit anything I would merge tomorrow or on Monday.

dyedgreen:
Sounds good 👍

mity:
Just noticed scripts/run-tests.sh passes outdated option `--flatex` instead of `--flatex-math`. Also perhaps keep the pathological tests as the last ones. Those are kind of different then all the other ones.

dyedgreen:
Should be fixed now. Sorry for that, there was a lot of back and forth 😅

mity:
No problem.

mity:
Playing with it right now a little bit and noticed this:

``` bash
$ printf '`a\`' | ./md2html/md2html --flatex-math
<p><code>a\</code></p>
$ printf '$a\$' | ./md2html/md2html --flatex-math
<p>$a$</p>
```
I.e. backslash in a code span and in the math span behaves differently. Is it intended?

(If not, parsing `$` would have to be done earlier before escape sequences; at the same time as code span. But it would require quite a big overhaul because those are detected during text scan in `md_collect_marks()`).

Also 
``` bash
$ printf '$*foo*$' | ./md2html/md2html --flatex-math
<p><equation><em>foo</em></equation></p>
```
Is that intended?
(If not, then using `md_rollback()` should use `MD_ROLLBACK_ALL` to disable all Markdown stuff inside the span; not just ranges crossing the opener or closer marks.)

dyedgreen:
> Playing with it right now a little bit and noticed this:
> 
> ```shell
> $ printf '`a\`' | ./md2html/md2html --flatex-math
> <p><code>a\</code></p>
> $ printf '$a\$' | ./md2html/md2html --flatex-math
> <p>$a$</p>
> ```
> 
> I.e. backslash in a code span and in the math span behaves differently. Is it intended?

This is correct, as it is an escaped `$`, which should not be interpreted as an equation delimiter. (There are also no valid LaTeX equations that end in `\`.)

> Also
> 
> ```shell
> $ printf '$*foo*$' | ./md2html/md2html --flatex-math
> <p><equation><em>foo</em></equation></p>
> ```
> 
> Is that intended?
> (If not, then using `md_rollback()` should use `MD_ROLLBACK_ALL` to disable all Markdown stuff inside the span; not just ranges crossing the opener or closer marks.)

That's wrong (although `*` is also unusual, hence why it slipped me). I'll update the code to do this correctly.

mity:
Also I wonder whether `$` should not be parsed with higher prioroty then most other Markdown stuff. Consider that math can contain quite a lot of chars which is otherwise meaningful to the Markdown parser.

E.g. `*` as multiply mark and not Markdown. emphasis.
Or `[]` as some matrix/vector indices or math parenthesis of larger importance then `()` and not Markdown link.
Or `<>` as a comparision sign instead of an autolink.
Etc.

This would mean to move `$` from the line 3894 and add it e.g. to 3861. It would still not fix it completely. The comment at 3859 is not true anymore: Some of the stuff (code spans, autolinks, raw HTML) has been recently moved directly to the `md_collect_marks()` as mentioned in my previous comment.

mity:
Do you have some set of more complex LaTeX math stuff which we could use to test by embedding on some Markdown paragraph which use extensively some links/emphasis etc. to see how to behaves?

dyedgreen:
> Also I wonder whether $ should not be parsed with higher prioroty then most other Markdown stuff. Consider that math can contain quite a lot of chars which is otherwise meaningful to the Markdown parser.

I think there are much fewer chars than you think (have you used LaTeX much before?). Most things are done via commands e.g.:
```LaTeX
\times -> *
\frac{a}{b} -> a/b
\left \( -> (
\right \) -> )
...
```
(for more examples see [here](https://oeis.org/wiki/List_of_LaTeX_mathematical_symbols) and [here](https://www.overleaf.com/learn/latex/Brackets_and_Parentheses))

That being said, it still might be useful, since some characters are very frequent like `_^\`. You are obviously much more familiar with the internal workings of the parser, so moving it to higher priority may be a sensible thing to do. (Although I'm not sure how that would have to be done)

dyedgreen:
> Do you have some set of more complex LaTeX math stuff which we could use to test by embedding on some Markdown paragraph which use extensively some links/emphasis etc. to see how to behaves?

[These guys have some tests](https://github.com/KaTeX/KaTeX/tree/master/test), although they are embedded into JavaScript documents. Another [set of tests is here](https://github.com/mathjax/MathJax/tree/master/test), this time embedded into html. I'm not sure how easy it would be to extract these.

mity:
> have you used LaTeX much before?

Many years ago and forgot it completely...

> ... moving it to higher priority may be a sensible thing to do. (Although I'm not sure how that would have to be done)

In short, by moving `$` from the line 3894 and add it e.g. to 3861. 

Longer answer is that earlier the given mark is handled, it has higher priority. Because in subsequent steps, the `md_analyze_marks()` **skips** contents of the resolved ranges. I.e. when it reaches **resolved** opener mark, the function skips to the corresponding closer mark and continues from there. (Only contents of links/images is later done again from scratch, sort of as a standalone paragraph, with few exceptions.)

If `$` would have lower priority, then e.g. `*`, then `$blah blah*$*` would become `$blah blah<em>$</em>`.

And if `$` and `*` have both the same priority as it is now, then it depends on what's resolved earlier by scanning the marks from left to right, as they appear in the text.

dyedgreen:
> In short, by moving `$` from the line 3894 and add it e.g. to 3861.
> 
> Longer answer is that earlier the given mark is handled, it has higher priority. Because in subsequent steps, the `md_analyze_marks()` **skips** contents of the resolved ranges. I.e. when it reaches **resolved** opener mark, the function skips to the corresponding closer mark and continues from there. (Only contents of links/images is later done again from scratch, sort of as a standalone paragraph, with few exceptions.)
> 
> If `$` would have lower priority, then e.g. `*`, then `$blah blah*$*` would become `$blah blah<em>$</em>`.
> 
> And if `$` and `*` have both the same priority as it is now, then it depends on what's resolved earlier by scanning the marks from left to right, as they appear in the text.

Right. So a very simple change. I still don't know if it is necessary though, from what I can tell, the kind of cases where the symbols mix (which is mostly `_<>`) are fine:
```Shell
$ printf '$a_b+c$ bla bla  _ bla bla' | ./md2html/md2html --flatex-math
<p><equation>a_b+c</equation> bla bla  _ bla bla</p>
$ printf '$a*b+c$ bla bla  * bla bla' | ./md2html/md2html --flatex-math
<p><equation>a*b+c</equation> bla bla  * bla bla</p>
$ printf '_ some text $a_b+c$ bla bla' | ./md2html/md2html --flatex-math
<p>_ some text <equation>a_b+c</equation> bla bla</p>
$ printf '< some text $a_b>c$ bla bla' | ./md2html/md2html --flatex-math
<p>&lt; some text <equation>a_b&gt;c</equation> bla bla</p>
```
(The last case should not escape the inner `>`, but that is done at the renderer level, and since the example render does not really support LaTeX, I don't see a problem there.)

mity:
Ok. If can leave that as it is now, and return to it if some problem is encountered.

dyedgreen:
Yes, that's my thinking too. Especially since this is an internal detail, it can be patched later if anyone discovers a nasty edge-case.

mity:
> (The last case should not escape the inner >, but that is done at the renderer level, and since the example render does not really support LaTeX, I don't see a problem there.)

HTML renderer **has** to escape it or it would be invalid HTML. An alternative would be instead to wrap the math in CDATA section without any escaping:


```
<equation><![CDATA[

...here we can now use anything, without the need of escaping...

]]></equation>

</body>
</html>
```

dyedgreen:
Yes. I probably wasn't very clear here. I meant that the md4c parser should not do any escaping here, so that a consumer of the api can correctly render the LaTeX. But since the escaping is done at the level of the HTML renderer it is correct.

mity:
Merged. Thanks for all the work.

dyedgreen:
Thank you for bearing with me through all the changes 😊

Issue #82, opened 2019-05-20
Making the HTML renderer part of public interface?


mity:
It seems there is some (growing?) demand to have the HTML renderer (implemented in `md2html/render_html.[hc]` + `md2html/entity.[hc]`) available for easy reuse. See e.g. #80 or https://www.reddit.com/r/rust/comments/b5cul0/benchmarking_markdown_libraries_comrak/ejg6c8m/

In some ways, I see it as quite a big step, especially if it becomes part of some "stable official API", and twice so if some Linux distros start to ship it and it becomes subject of some inter-package dependencies; and also because it could possibly make sort of a precedent for the future if some other renderers to more output formats perhaps get implemented.

So I have a lot of questions (and although I may have some opinions about some of them I am still open to discussion and interested in opinion of other people in different roles).

1. Should its code live in a new separate dir; or be moved into `md4c/`?
2. Should the renderer be built as a separate lib when configured to build as a **static** lib; or merged into `libmd4c.a`?
3. Should the renderer be built as a separate lib when configured to build as a **shared** lib; or merged into `libmd4c.so`?
4. Should it be instalable? (I.e. are distros or other projects interested in that?)
5. Should the header `render_html.h` be merged into `md4c.h` or live as a standalone?

In any case it should likely undergo some renaming (`render_html.h` or potential `librender_html.[a|so]` is likely not specific enough; the names should contain some "md4c" to mitigate name conflicts etc.), also its API should be reviewed and polished before making it officially public so do not expect this to happen instantly.

Feedback? Opinions? Complains?

(cc @perezmeyer @ec1oud @hvoigt)

perezmeyer:
Note: I'm understanding: provide the renderer as a library, not as an application. 

1. I would keep it where it is, or even split the test application into it's own directory.

2. For static builds I think that a separate lib makes it more clear that they are different things. Then the related .cmake config file should list *first* md4c and then libmd4crenderer as dependencies, so everything goes smooth (static linking should be done in the right order).

3. Same as above, dependencies will be handled automatically.

4. If you provide stable API I don't see why it shouldn't be installable.

5. Both would work, as the renderer will be highly dependant of the libmd4c version.

As for the naming: as I already wrote I think either libmd4crenderer is more clear and straightforward.

Hope that helps, feel free to ping me for more info/feedback if needed. And **thanks a lot** for asking!

hvoigt:
Here my 2 cents:

[...]

PR #89, opened 2019-08-09
Refactor: Build the HTML renderer as a standalone library

mity:
(Addresses #82)

**Current status:**

Right now I see it as a fairly complete, yet it is currently meant more as a proof of concept and material for further discussion. So feedback is welcome.

**Aims:**
1. Make the HTML renderer build as a standalone library (stable public API, pkgconfig .pc file etc.). 
2. Provide simpler interface for applications which need only the Markdown -> HTML conversion.
3. Should serve also as a model for renderers to other formats if/when those are implemented and added.

**Notes:**

1. It involves renaming the `render_html.[ch]` to `md_html.[ch]` and the identifiers from the header to make it more friendly/consistent as a public API.

2. It involves moving ~~all~~ **all potentially reusable** sources into a single (flat) `src` directory.

    Rationale: Generally I prefer one-target-per-directory structure, but it imho does not work/scale well here. If we add more renderers, each with one directory and containing (likely/in most cases) just one header and one source, yet they would still likely need to reuse the `entity.[ch]`.

3. Application should not need more then the parser flags from `md4c.h`. Strictly speaking, we might consider make it complete independent header and make the renderer translate its flags to the parser flags. 

   Disadvantage would be that then every (future) renderer whould then have do the same kind of translation for the sake of consistency.

   Advantage would be that renderers could implement or enforce/disable some flags if they do not/cannot implement some MD4C features and communicate it better in their API. For example I can imagine that most renderers (likely all but the HTML one) will never support the raw HTML blocks/inlines.

   I am not sure what is better here.


mity:
(Rebased the PR to make it up-to-date with the current master.)

mity:
(Rebased the PR on top of the current master.)

PR #92, opened 2019-09-09
Wiki links

niblo:
This does not feature link labels, only the link target. Comments appreciated.


md4c/md4c.h
```diff
@@ -268,6 +273,11 @@ typedef struct MD_SPAN_IMG_DETAIL {
     MD_ATTRIBUTE title;
 } MD_SPAN_IMG_DETAIL;
 
+/* Detailed info for MD_SPAN_WIKILINK. */
+typedef struct MD_SPAN_WIKILINK {
+    MD_ATTRIBUTE target;
+    MD_ATTRIBUTE text;
```
mity: The attribute `text` is IMHO not needed. It should rather be provided via `text()` callback as a normal text flow (and the way you update the renderer implementation suggests it already works that way).

md4c/md4c.c
```diff
@@ -3398,6 +3415,47 @@ md_resolve_links(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
...
+            (next_closer->beg == closer->beg + 1) &&
+            (next_opener->end - next_opener->beg == 1) &&
+            (next_closer->end - next_closer->beg == 1) &&
+            (next_opener->ch == '[' && next_closer->ch == ']'))
+        {
+
+            is_link = md_is_inline_wikilink_spec(ctx, lines, opener->end);
+            if (!is_link) {
+                opener_index = next_index;
+                continue;
```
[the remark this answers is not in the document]
niblo: You are right about that. Will fix.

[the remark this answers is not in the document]
niblo: Although I might add that with the current implementaiton of `md_is_inline_wikilink_spec`, I believe that what you mention is not possible.

md4c/md4c.c
```diff
@@ -2331,6 +2331,23 @@ md_is_inline_link_spec(MD_CTX* ctx, const MD_LINE* lines, int n_lines,
     return ret;
 }
 
+static int
```
mity: I also have a trouble to understand why this function is needed at all. The caller `md_resolve_links()` already gets position of opening `[` and closing `]`. So the rescanning of all the contents between is a work in a vain. And could even cause troubles if you analyze `[[ [[ ]] ]]` because the scanner in this function will stop too early.

md4c/md4c.c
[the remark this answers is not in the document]
mity: Please see my newly added comment about `md_resolve_links()`.

This test has to be done differently. You cannot check just the raw text via the `CH(off)` because that `[` might be escaped or already used.

I believe that we have to check `ctx->marks[]` in neighborhood of the opener and closer mark (not sure right now whether its outer or inner side) whether it is also `[` or `]` mark and whether it is not yet "resolved".

md4c/md4c.c
niblo: I tried to follow the pattern of `md_is_inline_link_spec`. In it, there is a check to make sure that the link target does not span two lines, which is the exact case used for wiki link targets. So, I thought it would be suitable. And also, when support for the wiki link label is added, this function may include further checks, like preventing the label from spanning too many lines.

md2html/md2html.c
```diff
@@ -209,6 +209,7 @@ static const option cmdline_options[] = {
     { "ftables",                     0,  'T', OPTION_ARG_NONE },
     { "fstrikethrough",              0,  'S', OPTION_ARG_NONE },
     { "flatex-math",                 0,  'L', OPTION_ARG_NONE },
+    { "fwiki-links",                 0,  'w', OPTION_ARG_NONE },
```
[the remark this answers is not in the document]
niblo: I will use `'K'` then.

md4c/md4c.c
mity: The (ordinary) link handling is one of the most difficult parts of MD4C. The complexity is to avoid non-linear behavior for crazy malicious inputs. 

That's why there is that craziness with the `md_resolve_links()` at all. Without it it would be simple to DoS it with
```
$ python -c 'N=20000; print("[" * N + "foo" + "]" * N);' | ./md2html
```

I'm afraid your scanning in this function introduces the quadratic complexity with `--fwiki-links` again, and also that it is likely broken anyway because if you feed it with something like

```
[[ foo[[ bar]]
]]
```

Then for the outer `[[...]]` you have the new line there anyway but the scanner only always scans to the 1st `]]`.

md4c/md4c.c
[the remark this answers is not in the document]
niblo: > Please see my newly added comment about `md_resolve_links()`.
> 
> This test has to be done differently. You cannot check just the raw text via the `CH(off)` because that `[` might be escaped or already used.

Can you think of a test case for this? I am trying to come up with something that would fail in a test, but I have no luck. I think that the giant if-statement that does the wiki link detection prevents such a case from happening.

md4c/md4c.c
niblo: > The (ordinary) link handling is one of the most difficult parts of MD4C. The complexity is to avoid non-linear behavior for crazy malicious inputs.
> 
> That's why there is that craziness with the `md_resolve_links()` at all. Without it it would be simple to DoS it with
> 
> ```
> $ python -c 'N=20000; print("[" * N + "foo" + "]" * N);' | ./md2html
> ```

I see.

> I'm afraid your scanning in this function introduces the quadratic complexity with `--fwiki-links` again...

That's unfortunate. I will have another look at it and see if I can do something about it.

> ... and also that it is likely broken anyway because if you feed it with something like
> 
> ```
> [[foo[[bar]]
> ]]
> ```
> 
> Then for the outer `[[...]]` you have the new line there anyway but the scanner only always scans to the 1st `]]`.

This is the output that I get:

```
<p>[[foo<x-wikilink data-target="bar">bar</x-wikilink>
]]</p>
```

But that is what I expect:

- Links (and wiki links) are scanned from inside to out,
- Wiki link target cannot span more than one line

niblo: > ... and also that it is likely broken anyway because if you feed it with something like
> ```
> [[foo[[bar]]
> ]]
> ```
> 
> Then for the outer `[[...]]` you have the new line there anyway but the scanner only always scans to the 1st `]]`.

Excuse me; maybe I was a bit too quick in my previous response. Do you mean that the following should be expected?
 
```
<p><x-wikilink data-target="foo[[bar">foo[[bar</x-wikilink>
]]</p>
```

mity: I think I have a proposal which would fix all of that:

Change the `md_is_inline_wikilink_spec` in the follow way:
1. Make it take `beg` and `end` offsets from the opening/closing `MD_MARK` so we really use the proper marks.
2. Make it scan backwards from `end`. It will come handy later when implementing the pipe case (we can then only scan over the ID, and keep the text without length limit).
3. Make it scan until it reaches `beg` instead of checking for `[[` because there may be nested `[[ ... ]]`.
4. Make it fail if it sees a forbidden char (`'\r'` or `'\n'` - here, it should be fine to use instead of analyzing the `lines[]`).
5. Make it scan at most, say 100 characters, and fail if it does not reach `beg`. (I think we can limit the ID length right?)

And I would also rename the function to `md_is_inline_wikilink_id()`. The `_spec()` for ordinary link is used because it is for the whole link destination + link title pair.

md4c/md4c.c
```diff
@@ -3398,6 +3415,47 @@ md_resolve_links(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
...
+            if (!is_link) {
+                opener_index = next_index;
+                continue;
+            }
+
+            opener->beg = next_opener->beg;
+            closer->end = next_closer->end;
+
+            /* This does not seem to do much: */
+            /* next_opener->ch = 'D'; */
```
niblo: Do you have some comment on this? There is something that I'm not understanding. I was expecting the wiki link detection not to work without it, but it does work.

md4c/md4c.c
mity: > Excuse me; maybe I was a bit too quick in my previous response. Do you mean that the following should be expected?

Not necessarily. 

> But that is what I expect...

Maybe I constructed bad input exaple. 

The culprit of my message is that `md_resolve_links()` is called for each pair of `[` and `]` (following the nesting logic that inside you may get only balanced pairs), but your scanner will stop on the first `]]` so it has **a different logic** and does not necessarily scans whole `[[...]]` range as analyzed in the `md_resolve_links()`.

md4c/md4c.c
mity: I believed it is needed for some obscure inputs like `[[foo]](bar)`, thinking it goes roughly as follows:

1. `md_analyze_links()` is called for the inner `[...]` first. 
2. You check whether it is enclosed in another `[...]`. It is -> hence you resolve as a wikilink.
3. Then `md_analyze_links()` is called for the outer `[...]`.
4. It sees it makes an inline links (with the `bar` as its destination). The link demolishes everything detected inside of it and then analyzes its text as sort of a standalone paragraph from scratch, killing the previous wiki-link detection.

Clearly, there is something else what makes it all play differently. Maybe I overlooked/forgot something... Needs more analysis.

md4c/md4c.c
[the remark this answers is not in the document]
mity: > Can you think of a test case for this?

Took me a while but yes:
```
$ printf '[[ ![foo\]](url) \n]]' | ./md2html/md2html.exe --fwiki-links
<p><x-wikilink data-target=" ![foo]](url)
"> <img src="url" alt="foo]">
</x-wikilink></p>
```

The trick is to use image because image can be nested in a link while link a link cannot be nested. Here your scanner stops on `\\]]` which makes no sense at all. And its real end is on the next line.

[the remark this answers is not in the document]
mity: Well, thinking more about it, it is sort of the same objection as the `md_is_inline_wikilink_spec()` inconsistency. Another counter-example is this:

```
[[`]]`
]]
```

I.e. hiding the false `]]` into a code span as a trap for the `md_is_inline_wikilink_spec()`.

md4c/md4c.c
niblo: > I think I have a proposal which would fix all of that:
> 
> Change the `md_is_inline_wikilink_spec` in the follow way:
> 
>     1. Make it take `beg` and `end` offsets from the opening/closing `MD_MARK` so we really use the proper marks.
> 
>     2. Make it scan backwards from `end`. It will come handy later when implementing the pipe case (we can then only scan over the ID, and keep the text without length limit).
> 
>     3. Make it scan until it reaches `beg` instead of checking for `[[` because there may be nested `[[ ... ]]`.
> 
>     4. Make it fail if it sees a forbidden char (`'\r'` or `'\n'` - here, it should be fine to use instead of analyzing the `lines[]`).
> 
>     5. Make it scan at most, say 100 characters, and fail if it does not reach `beg`. (I think we can limit the ID length right?)
> 
> 
> And I would also rename the function to `md_is_inline_wikilink_id()`. The `_spec()` for ordinary link is used because it is for the whole link destination + link title pair.

I did this; the result is much simpler. Thanks for the good advice.

md4c/md4c.c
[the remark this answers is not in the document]
niblo: With the re-implementation `md_is_inline_wikilink_id`, I get the following from your first test case here:

```
<p>[[ <img src="url" alt="foo]">
]]</p>
```

That looks right, doesn't it? And the second:

```
<p>[[<code>]]</code>
]]</p>
```

Do you think they should be included as test cases?

md4c/md4c.c
mity: Seems to fix my counter-examples from above so seems to be on the right track now.

md4c/md4c.c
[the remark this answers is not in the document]
mity: Probably no.

mity:
Seems good to me now so far. Do you plan to continue to work in the foreseeable future in this PR?

(It would be simpler for me, as such a feature deserves proper testing, including some fuzzing over-night so I would wait with that until you consider it complete to do it just once.)

niblo:
> Seems good to me now so far. Do you plan to continue to work in the foreseeable future in this PR?

I will continue with adding support for the label part of the wiki link during the week. The goal is to complete the wiki link feature, with proper support for labels.

test/wiki-links.txt
```diff
@@ -37,3 +37,48 @@ Wiki links are prioritized over regular links.
...
+.
+<p>[<a href="/url">foo</a></p>
+````````````````````````````````
+
+Escaping the opening `[`, we should get a normal link.
+
+```````````````````````````````` example
+\[[foo]]
+.
+<p>[<a href="/url">foo</a>]</p>
```
niblo: By the way, here is a failing test case. I'm not sure what should be the proper behavior here. I was hoping you had some input.

mity:
Ok. I would then fuzz-test it together when it is done. 

It takes AFL on my machine many hours to cover all the code paths and it is almost unusable for anything else in the mean time. Hopefully will qualify to https://github.com/google/oss-fuzz one day but until then...

test/wiki-links.txt
mity: That `[foo]` inside is not a link because it has no link destination and there is no corresponding link reference definition, so it renders correcly to `<p>[[foo]]</p>`.

If I add the definition:
```
[foo]: /url

\[[foo]]
```

then it outputs what you expect.

md4c/md4c.c
mity: I think the link nesting rules save it here. CommonMark specifies the links cannot be nested in each other (and if they are, only the innermost is really a link). This is handled in `md_analyze_links()` with this piece of code:

``` C
        /* If nested ("[ [ ] ]"), we need to make sure that:
         *   - The outer does not end inside of (...) belonging to the inner.
         *   - The outer cannot be link if the inner is link (i.e. not image).
         *
         * (Note we here analyze from inner to outer as the marks are ordered
         * by closer->beg.)
         */
        if((opener->beg < last_link_beg  &&  closer->end < last_link_end)  ||
           (opener->beg < last_img_beg  &&  closer->end < last_img_end)  ||
           (opener->beg < last_link_end  &&  opener->ch == '['))
        {
            opener_index = next_index;
            continue;
        }
```

Therefore the outer ones are not really considered.

mity: But it is a different story with an image:
```
![[foo]](url)
```
translates into
```
<p><img src="url" alt="[foo]"></p>
```

Here the reason is that when handling the inner pair in `md_analyze_links()`, the `next_opener` points `MD_MARK` describing the image opener `![` and it has `next_opener->ch == '!'` therefore the big `if` for the wiki-link detection fails.

The question is what precedence rules we ant to here.

Anyway, I would remove those two lines, even if we decode to change the precedence to prefer wiki-link here, it would have to be done differently.

md4c/md4c.c
```diff
@@ -3497,6 +3535,74 @@ md_resolve_links(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
...
+
+    /* No content; ignore. */
+    if (opener->end == closer->beg)
+        goto not_found;
+
+    while(delim_index < closer_index) {
+
+        /* The wiki link is ignored if the `|` delimiter is used, but without a
+         * label. TODO Unsure if this is the desired behavior. How about
+         * removing the `|` from the output instead, and keeping the link? */
```
mity: In general, I would follow what other important wiki implementations (wikipedia?) usually do.

md4c/md4c.c
```diff
@@ -3497,6 +3535,74 @@ md_resolve_links(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
...
+         * a target. */
+
+        if(delim->ch == '|' && (delim->end == closer->beg || delim->beg == opener->end)) {
+            goto not_found;
+        } else if(delim->ch == '|') {
+            opener->end = delim->beg;
+            break;
+        }
+
+        delim_index++;
```
mity: This scanning for the `|` likely opens the gates to quadratic behavior for `[[[[[[[[foo|bar]]]]]]]]....`.

md4c/md4c.c
```diff
@@ -3497,6 +3535,74 @@ md_resolve_links(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
     return 0;
 }
 
+static int
```
mity: I have some trouble understanding why this big function is needed. Even if it is needed it has some issues I commented in its body.

Generally, I assumed we can still scan only backwards from the `]]` as before. The only novelty would be that when we reach a `|`, we use it as a beginning of the ID from that point (the closing `MD_MARK` would be expanded to hide it from normal text flow), and that's it.

Also note we then do not have to collect `|` marks beforehand for wiki-links: We reach that char by the scanning over the ID anyway, so it would simply the logic quite a lot.

md4c/md4c.c
```diff
@@ -3497,6 +3535,74 @@ md_resolve_links(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
...
+        delim_index++;
+        delim = &ctx->marks[delim_index];
+    }
+
+    OFF off = closer->end;
+    int count = 0;
+    int newlines = 0;
+    int has_label = (opener->end - opener->beg > 2);
+
+    /* Check for newlines. */
```
mity: I do not see a point in this at all. The block parsing (the1 st parsing big step) already guarantees we do not try to match `[[` with `]]` across blocks (i.e. there is no blank line for example), and I do not see a principal reason why 
```
[[a long text
over multiple
lines|ID]]
```
should be rejected.

So imho we can still happily only scan over the ID part.

Also your approach would not be reliable:
1. `"\r\n"` is considered one new line digraph in Markdown.
2. Once you do not simply fail when reaching a new line, you can happily scan over contents which does not belong to the block contents like block decorations or indentation, breaking the logic with the text length limit.

    Consider e.g. a multi-line wiki-link used in a nested blockquote:
    ```
    > > > > [[foo
    > > > > bar|id]
    ```

md4c/md4c.c
niblo: > Generally, I assumed we can still scan only backwards from the `]]` as before. The only novelty would be that when we reach a `|`, we use it as a beginning of the ID from that point (the closing `MD_MARK` would be expanded to hide it from normal text flow), and that's it.

The *closing* mark? Don't you mean the opening mark?

> Also note we then do not have to collect `|` marks beforehand for wiki-links: We reach that char by the scanning over the ID anyway, so it would simply the logic quite a lot.

That would be neat.

mity: > The closing mark? Don't you mean the opening mark?

`[[text|id]]`... I.e. the closer mark would be expanded back to cover not only `]]` but whole `|id]]`.

niblo: Aha, then it would be so, if not for the fact that the wiki link structure is the other way around: `[[id|text]]`.

mity: Oops. Mea culpa. So I got it whole wrong and we also need scanning forward from the opener, not backward. Whatever. Does not change the principle.

niblo: > > Also note we then do not have to collect `|` marks beforehand for wiki-links: We reach that char by the scanning over the ID anyway, so it would simply the logic quite a lot.

Can I get the `MD_MARK` for the `|` by way of it's position in the text somehow?

niblo: The `|` can be escaped inside the wiki link, so that has to be accounted for. Any pointers?

mity: > Can I get the `MD_MARK` for the `|` by way of it's position in the text somehow?

There is no infrastructure for it and it would be better to avoid such need. 

Even doing some binary search would be complicated because although the marks are originally ordered by their natural positions, during the resolving they may be played with and that breaks that property (e.g. they may be collapsed to a zero-length, expanded, there are the dummy marks which actually are only containers for other data so their `beg` and `end` members are misused for other purposes etc.

> The `|` can be escaped inside the wiki link, so that has to be accounted for. Any pointers?

I would likely ignore the marks for that purpose here altogether, it would make it just too complicated. I would simply count the number of `\` just before the pipe (ideally on the fly during the scanning we already do). If it is an odd number, it is escaped.

niblo: Do you have any handy tips on marking the `|` as resolved?

niblo: The `|` is still appearing in the output, so I assume I must mark it as resolved somehow. Is there a suitable way to do that in this context, or is there something else I should do?

mity: Ouch. A weak spot in my consideration. 

I have an idea, but I am not sure whether it would really work (I do not have all the stuff in my head any more), so take the following with a little bit of salt.

`md_collect_marks()` already places a dummy `MD_MARK` after every `[` because if it gets later resolves as a link we need a space for storing some data.

But if I remember correctly, it is not used at all unless the link is eventually resolved (if it happens at all). So technically we could store some data there which could later be retrieved in the order to get the index of the `|` mark in the `ctx->marks[]`.

I.e. something like this:
1. Make `md_collect_marks()` alweas remember the index of last such dummy after the most recent `[`.
2. When `md_collect_marks()` encounters `|` and it constructs its corresponding `MD_MARK`, we could store its index in the remembered dummy mark (if it is not already set because we want only to register the 1st such pipe).
3. When you are resolving the `[[...]]`, you should be able to dig it from there and verify whether it is between the `[[` and `]]` (and not somewhere later).

It could then also be used for the escaped pipe detection because `md_collect_marks()` escapes escaped characters and does not create `MD_MARK` for escaped chars at all.

mity: > The `|` is still appearing in the output, so I assume I must mark it as resolved somehow. Is there a suitable way to do that in this context, or is there something else I should do?

Uff? You may hide it be expanding the opening `[[` to cover not just the ID but also the pipe char.

But we likely still need to do something with the mark in order to hide it from a table cell handling. But not resolving (resolving means, the `md_process_inlines()` would make a stop there to consider calling `enter_span()/leave_span()` callback). I would change it to dummy instead.

mity: I have just realized the above suggestion wouldn't work. We do not have any guaranty the pipe has to belong to the most recent `[`.

Counter-example is this:
```
[[[id_in_brackets]|text]
```

niblo: To sum up, the initial attempt at label handling works, but the following has not been considered:

- possible quadratic behavior (although I cant find this)
- leading quote block marks are counted in the wiki link content limit

Also, you think that there should not be a limit on how many lines the label may span, and I agree with this.

For counting the number of leading characters that make up the block quote, I found some interesting code in the beginning of `md_analyze_line`. Do you think it would be suitable for use here?

And, do you still think the wiki link routine should be inlined in `md_resolve_links`, or could it be contained in its own function?

niblo: Continuing from where I left off, it looks like `total_indent` as used in `md_analyze_line` has exactly the information that is needed here.

niblo: One solution is to hand over `total_indent` to the `MD_LINE` that is created in `md_add_line_into_current_block`, but that would mean one more member to `MD_LINE`/`MD_LINE_tag`.

mity: > For counting the number of leading characters that make up the block quote, I found some interesting code in the beginning of md_analyze_line. Do you think it would be suitable for use here?
> ...
> Continuing from where I left off, it looks like `total_indent` as used in `md_analyze_line` has exactly the information that is needed here.

No way. This would break everything and it would make the complexity problem worse.

1. It is the main part for detecting starts and ends of blocks, determining their type from line prefix/indentation. The problem is it cannot be reasonably used anywhere else because it inputs needs the type of previous line and also it iterates over arbitrary amount of characters and there is no limit on line length. Would we call it since beginning of document until we reach the line we are interested in? I cannot imagine that.

2. The function `md_analyze_line()` is super expensive. It scans over every char in the line.

It's worse then scanning some part of it manually. If we would hypothetically call `md_analyze_line()` just once per a wiki-link, we have a qudrattic behavior if you please big number of wiki-links to a single enormously long line.

3. Also there are special cases where it would fail horribly, e.g. when nested in something what looks like a link or image:
```
![  [[wiki]] ](/img/url)
```
because in those cases, we artificially create vector of `MD_LINE`s for contents of the image and process its contents partially as a standalone paragraph.

> One solution is to hand over `total_indent` to the `MD_LINE`.

1. It would extremely complicate things e.g. for the image nesting as above or for tables, as there we artificially construct `MD_LINE` too for each table cell.

2. This is the most critical structure for memory overhead of MD4C. We construct a vector of blocks and all their `MD_LINE`s per whole document before the inline processing. So 3 integers instead of 2 in this struct is a big change. I certainly won't do anything like that with a light heart.

mity: > * possible quadratic behavior (although I cant find this)

I spent some thinking about it.

Maybe it is not there, at least if the iterating over `MD_MARK`s is done very carefully: If the ID length has a limit (say 100 characters), then it also limits the number of potential `MD_MARK` we have to iterate in order to find the pipe: If we reach any (non-dummy) `MD_MARK` which has `MD_MARK::beg > opener.end + 100`, we know there is no such pipe mark.

We must be careful to now allow iteration over arbitrary (unlimited) stuff. But if we can guarantee it has a (reasonable) limit, it should be ok.

(But sorry, I don't have time today to dig in the code today to verify whether it works that way already. Busy with other things.)

niblo: I didn't mean to make a call to `md_analyze_line`. What I meant was that `total_indent` could be saved in the `MD_LINE_ANALYSIS` struct, and then handed/copied over to the `MD_LINE` struct in `md_add_line_into_current_block`. That way, in `md_resolve_links`, because we have access to the line(s), we would have access to the `total_indent` value.

mity:
@niblo Sorry for the long delay. Solving some real-life stuff. 
I hope to get to it again during the week-end.

niblo:
Sounds good.

mity:
Fuzz testing has found many crash(es) in this branch, but likely they are all exhibitions of the same bug (multiple randomly chosen crash with the same backtrace). 

Looks to me that there its some interference of code for tables and for wiki-links as it crashes only with `--ftables`. It even does not need `--fwiki-links` but the crash does not happen on master branch.

I'll try to minimize and textize the crashing input (it has tens of kB of ugly binary stuff) before posting it here.

mity:
Minimized one of the crash inputs into this:

```
[f]: &#ffff;

x|x
---|---
[foo|x]

x|x
---|---
[foo|x]
```

md4c/md4c.c
```diff
@@ -3236,8 +3238,8 @@ md_collect_marks(MD_CTX* ctx, const MD_LINE* lines, int n_lines, int table_mode)
                 continue;
             }
 
-            /* A potential table cell boundary. */
-            if(table_mode  &&  ch == _T('|')) {
+            /* A potential table cell boundary or wiki link label delimiter. */
+            if((table_mode || ctx->parser.flags & MD_FLAG_WIKILINKS) && ch == _T('|')) {
```
mity: Noticed this issue in the code: Bit operator `&` has higher priority then `||` so the part of the condition in the brackets is equivalent to `(table_mode | flags) & WIKILINKS`. It should be `table_mode | (flags & WIKILINKS)`.

mity: Sorry. Not true, it's `==` which usually makes troubles. But I would prefer if you add the parenthesis there anyway, please.

niblo:
Thanks for the feedback. I will try to have a look at it this weekend.

md4c/md4c.c
```diff
@@ -3872,14 +3968,6 @@ md_analyze_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines, int table_mod
         return ret;
     }
 
-    /* (3) Links. */
-    md_analyze_marks(ctx, lines, n_lines, 0, ctx->n_marks, _T("[]!"));
-    MD_CHECK(md_resolve_links(ctx, lines, n_lines));
-    BRACKET_OPENERS.head = -1;
-    BRACKET_OPENERS.tail = -1;
-    ctx->unresolved_link_head = -1;
-    ctx->unresolved_link_tail = -1;
```
niblo: It seems like the issue goes away if I revert this, so that links are analyzed before tables.

niblo: My mistake. It's still there, but the crash/bad output does not happen all the time, and I did not run md2html with the bad input enough times.

niblo: No, wait. It actually fixes the issue. I was jumping back and forth between branches. What I meant was: If the links analysis is moved to *after* the table analysis, it fixes the issue. So I will attempt to dig a bit deeper into this. Sorry for the confusion.

niblo: Update: Analyzing links before tables produces the same crash on the master branch too.

@mity: Can you tell if I did something stupid when moving the link analysis to occur before the table analysis? That would be in `md_analyze_inlines`.

niblo: As a sidenote, I seem to have found a related issue. This is on the master branch with no changes on top.

----
```
[a]: xyz

[abc]
```
```
<p><a href="xyz">abc</a></p>
```
----
```
[abc]: xyz

[a]
```
```
<p><a href="xyz">a</a></p>
```
----
```
[abc]: ijk
[a]: xyz

[a]
```
```
<p><a href="xyz">a</a></p>
```
----
```
[abb]: ijk
[a]: xyz

[a]
```
```
<p><a href="ijk">a</a></p>
```
----
```
[a]: ijk
[abc]: xyz

[abc]
```
```
<p><a href="xyz">abc</a></p>
```
----
```
[a]: ijk
[abb]: xyz

[abb]
```
```
<p><a href="ijk">abb</a></p>
```

mity: > It seems like the issue goes away if I revert this, so that links are analyzed before tables. Will dig a bit deeper into this...

Interesting.

> As a sidenote, I seem to have found a related issue. This is on the master branch with no changes on top.

Good catch. Will take a look at it.

mity: I have fixed the link label comparison in master. With that applied, you need another an updated test-case for the crash. This "works" for me:

```
[a|b]: /url

x|x
---|---
[a|b]
```

niblo: Another clue: the body of `if (is_link)` in `md_resolve_links` is executed twice when the crash/bad output appears, but only once when it runs without errors.

niblo: Another clue: in `md_process_table_row`, `n` is at some point 0 when md2html is called with the crashing input. This may result in `malloc` returning NULL (which is considered), but it may also result in an address that should not be dereferenced, according to my understanding. The latter case is not taken into consideration it seems.

mity: Playing in debugger with it a little bit.

I can see that with this input
```
[a|b]: /url

x|x
---|---
`a|b`
```

`md_is_table_row()` returns `FALSE` for the last line. That is fine because the pipe is hidden inside the code span.

But if it is the link as in
```
[a|b]: /url

x|x
---|---
[a|b]
```

then `md_is_table_row()` returns `TRUE`. That is wrong. Maybe this comes from some processing of the link contents because unlike for code spans, links may contain nested inline syntax elements so their contents is recursively re-analyzed from `md_resolve_links()` similarly as a normal paragraph. But that's just pure speculation at this moment.

Anyway, disabling this link contents processing when called in the context of `md_is_table_row()` is imho desirable even if it would not cause the issue: it is clearly an unneeded work. In this context we just need to see whether there is any pipe which marks cell boundary; i.e. any pipe which is _not_ inside any code span or link or anything else with higher priority.

mity: Just seeing your two most recent comments. I would say those are quite likely just consequences of `md_is_table_row()` getting bad answer in the 1st place.

niblo: I am not sure if I understand correctly, but I tested disabling the link analyzing and resolving in `md_analyze_inline` when it is called from `md_is_table_row`, but it does not solve the issue.

mity: I see, maybe the culprit is elsewhere in `md_is_table_row()`. But it indeed **should** return `FALSE`. The pipe is inside the link body, links have higher priority and therefore it is not usable as a table cell boundary.

niblo: `md_is_link_reference` is returning false when it should be true in `md_resolve_links`. It is the call that is commented with `/* Might be collapsed reference link */`.

mity: Ouch. Ouch. Ouch. That makes it all clear to me. And it points to an aspect which may make all the idea of links having higher priority then tables very problematic. We have a very ugly catch-22 here.

The problem is that `md_is_table_row()` is called when we are still determining whether the line belongs to a table block. To detect where the table ends and another block starts. I.e. during the very 1st pass for block analysis. Yes, `md_is_table_row()` called some inline analysis of the given line to resolve the higher precedence of some stuff like code spans, escaped pipes over the pipes used so the code spans in the table are possible and it also elegantly solves the case of escaped pipes.

The issue with links is that they can only be really resolved when complete block analysis is over and we know **all** link reference definitions in the document. (Because of this, the hash-table is currently only populated after we know all of them and that's why we hit the issue even if the ref. def. is before.)

mity: Interesting, I have just tried to do some experiments with github's Markdown. They clearly do not stop the table when there is a line without pipe.

```
A | B
--- | ---
no-pipe-here-but-still-a-table-row
```

renders into 

A | B
--- | ---
no-pipe-here-but-still-a-table-row

Hence, they likely do not need any line analysis at that early time. In that case I think we can change table parsing in a similar way. Improving the compatibility with them is a good thing on its own.

If I am right (and I believe I am) it should then fix the crash as a side effect. Actually it might even simplify quite a lot some things around the table parsing, maybe even remove the `md_is_table_row()` completely.

 I'll try to cook something (in another branch because it is quite independent from this PR.)

niblo: That sounds like a clever workaround if it works. Looking forward to seeing if it works.

mity:
I have just experimentally merged this PR and PR #97 and it seems to fix the crash. So, I'll fuzz-test that over the night and if there is no other surprise I'll then merge both tomorrow.

mity:
Finally merged.

@niblo, thanks for all the work and also for your patience.

niblo:
It was my pleasure. I would like to thank you too for the very same reasons. Your support made it possible.

Issue #101, opened 2020-01-10
Feature request: support for underlines


Waqar144:
Hello,

Thank you for this great library. We are [switching](https://github.com/pbek/QOwnNotes/pull/1563) to MD4C already in [QOwnNotes](https://github.com/pbek/QOwnNotes)

We were using hoedown before and it had optional support for underlines i.e., The following text would be rendered as underline

```md
_Underlined text_
```

Perhaps you can add this feature to MD4C. Maybe by using a flag such as `MD_EMPHASIC_CHAR_UNDERSCORE` and then using that to decide whether to render the text as underlined or italic.

mity:
If we do that, we should likely be more or less compatible when the flag is used.

So it would be good to know how this differs from the current parsing and whether it is "just" about interpretation (i.e. to see `_` as underline instead of (strong) emphasis) or whether it has also to be parsed differently.

Could you please check how hoedown behaves in these cases?

1. Two (or more) underscores: `__foo__`
2. Non-matching count of underscores: `__foo_`
3. Intra-word underscores: `foo_bar_baz`

Waqar144:
> Could you please check how hoedown behaves in these cases?
> 
>     1. Two (or more) underscores: `__foo__`

It renders this as **foo** [BOLD]

>     2. Non-matching count of underscores: `__foo_`

This gets rendered as `_foo(underlined)`

>     3. Intra-word underscores: `foo_bar_baz`

This gets rendered as `foo|bar(underlined)|baz`. 

Picture:

![image](https://user-images.githubusercontent.com/7696024/72155722-8a537f80-33d5-11ea-8c91-5984abbc1db1.png)


----------------

This isn't supported by CommonMark. In my opinion, `underscore` should just be interpreted the same way as it is already is being interpreted(as specified by the commonmark spec). Only when converting to html, the underscore is (optionally) rendered as underline.

mity:
Thank you. I will take a look at it.

mity:
> underscore should just be interpreted the same way as it is already is being interpreted(as specified by the commonmark spec). Only when converting to html, the underscore is (optionally) rendered as underline.

Hmm. I'm wondering whether it is a good idea, given how and why CommonMark does the emphasis.

For example consider three underscores:
````
___foo___
````

It now renders (as requested by CommonMark) into:
```
<p><em><strong>foo</strong></em></p>
```

That makes sense, given the `<em>` and `<strong>` distinction.

But if we render 
```
[...]

PR #103, opened 2020-01-10
Implement an underline extension.

mity:
(to address #101)

The rules to decide when we see `_` as a special char or an ordinary char are the same as in CommonMark. But with `MD_FLAG_UNDERLINE`, every single `_` seen as a valid is then translated into `<u>` (or `</u>`).

This effectively means that `__foo__` without the flag is rendered into `<strong>foo</strong>`, but with the flag enabled it becomes `<u><u>foo</u></u>`.

@pbek @Waqar144 Are you fine with it this way?

For testing, `md2html` now understands the command line option `--funderline`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ## Next Version (Work in Progress)
 
+New features:
+
+ * With `MD_FLAG_UNDERLINE`, spans enclosed in underscore (`_foo_`) are seen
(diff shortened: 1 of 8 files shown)


Waqar144:
WOW! :rocket: :rocket: 

Thank you so much.

Its looking like this:
image


Really great. Handles all the inputs as expected!

PR #115, opened 2020-05-29
Comment out unused parameters

rlohning:
I only checked whether they _are_ used,
not whether they _should be_ used.


rlohning:
Not allowed in C, I see.

PR #116, opened 2020-05-29
Add missing <img /> tag for XHTML support

dominickpastore:
While updating PyMD4C to support the new `MD_HTML_FLAG_XHTML` option, I noticed the `` tag was left unchanged. This small PR makes the `` tag self-close.


mity:
Good catch.

Thank you.

PR #119, opened 2020-06-28
Add .gitignore

cxw42:
Thank you for this project, which I have just started using and which seems very useful!  I noticed that compiling md4c dirties the Git repo, so I added a .gitignore file.  I submit it here in hopes it will be helpful to you or others.  Thanks for considering this PR!

Note on sources: the .gitignore is mostly the [GitHub C gitignore](https://github.com/github/gitignore/blob/master/C.gitignore).  I have added some lines for autotools, since I am using autotools to build a project that uses md4c as a submodule.  

Notes on contents: `Makefile` is ignored.  I think this is OK since you are using CMake instead of make, but I am happy to change it if you prefer.  I can also add `/build/` if you want that, since that's the CMake build dir for Travis.  

Please let me know if you have any questions or comments.  Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,78 @@
+# Prerequisites
+*.d
+
+# Object files
+*.o
+*.ko
+*.obj
+*.elf
+
+# Linker output
+*.ilk
+*.map
+*.exp
+
+# Precompiled Headers
+*.gch
+*.pch
+
+# Libraries
+*.lib
+*.a
+*.la
+*.lo
+
+# Shared objects (inc. Windows DLLs)
+*.dll
+*.so
+*.so.*
+*.dylib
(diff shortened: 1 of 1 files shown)


mity:
> Notes on contents: Makefile is ignored. I think this is OK since you are using CMake instead of make, but I am happy to change it if you prefer. 

That's ok, as Makefile is generated (if people use the CMake's make generator).

> I can also add /build/ if you want that, since that's the CMake build dir for Travis.

I don't think Travis cares so I'm indifferent. It can always be added later when a need arises.

mity:
Merged. Thanks.

cxw42:
Thanks very much for the quick review and approval!

PR #120, opened 2020-07-12
Fix a comment typo in md4c-html.h, md_render_html -> md_html

eklitzke:


mity:
Oh yes, it's a relic of a past code refactorization. Thanks for it.

PR #127, opened 2020-08-19
Update md4c.h

karstenBriksoft:
there was a typo in md4c.h: **Not** that in order to use it, you have to define
I've changed that to **Note**


mity:
Merged, thanks.

PR #128, opened 2020-09-12
FIx unitialized variable

jschueller:


mity:
Thanks for it but I believe all those variables actually are initialized. Check the `MD_CHECK` macro and its use in those functions. So I don't currently see any reason for merging this PR, at least in its current state.

That said, however, your PR points a finger to a real problem: In retrospect, I agree the code may be confusing as it is now and it should be fixed. I would say, we likely should get rid of that macro and do simply what it does manually at each call site, to improve the readability of the code.

jschueller:
Its mainly to silence the warning, I dont think there's anythong wrong, but it doesnt hurt.

PR #132, opened 2020-10-18
Strengthen build

rsms:
This set of changes improves some minor things, allowing compilation with stronger warnings (`-Wextra`) to reduce risk of undefined behavior. This patch also removes a few unused parameters which should yield smaller and maybe faster generated code. It also fixes an uninitialized variable (in `md_analyze_emph`) and partially initialized struct (`md_dummy_blank_line`).

In addition this contains a build script I find useful myself, which I've used in other projects. It's added to the `scripts` directory and is completely optional.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ dkms.conf
 # autotools
 /aclocal.m4
 /autom4te.cache/
+/build/
 /build-aux/
 conf*.dir/
 /config.h
--- a/md2html/md2html.c
+++ b/md2html/md2html.c
@@ -44,6 +44,12 @@ static int want_fullhtml = 0;
 static int want_xhtml = 0;
 static int want_stat = 0;
 
+/* MD_UNUSED macro for telling the compiler about intentionally unused arguments */
+#if defined(__has_attribute) && __has_attribute(unused)
+    #define MD_UNUSED __attribute__((unused))
+#else
+    #define MD_UNUSED
+#endif
 
 /*********************************
  ***  Simple grow-able buffer  ***
@@ -288,7 +294,7 @@ static const char* input_path = NULL;
 static const char* output_path = NULL;
 
 static int
-cmdline_callback(int opt, char const* value, void* data)
+cmdline_callback(int opt, char const* value, MD_UNUSED void* data)
 {
     switch(opt) {
         case 0:
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -3683,7 +3683,7 @@ md_analyze_emph(MD_CTX* ctx, int mark_index)
     /* If we can be a closer, try to resolve with the preceding opener. */
     if(mark->flags & MD_MARK_POTENTIAL_CLOSER) {
         MD_MARK* opener = NULL;
-        int opener_index;
+        int opener_index = 0;
 
         if(mark->ch == _T('*')) {
             MD_MARKCHAIN* opener_chains[6];
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -472,7 +472,7 @@ md_text_with_null_replacement(MD_CTX* ctx, MD_TEXTTYPE type, const CHAR* str, SZ
 typedef struct MD_UNICODE_FOLD_INFO_tag MD_UNICODE_FOLD_INFO;
 struct MD_UNICODE_FOLD_INFO_tag {
     unsigned codepoints[3];
-    int n_codepoints;
+    unsigned n_codepoints;
 };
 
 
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -62,6 +62,7 @@ dkms.conf
 # autotools
 /aclocal.m4
 /autom4te.cache/
+/build/
```
rsms, line 65: Since https://github.com/mity/md4c/wiki/Building-MD4C suggests using this directory

md2html/md2html.c
```diff
@@ -288,7 +294,7 @@ static const char* input_path = NULL;
 static const char* output_path = NULL;
 
 static int
-cmdline_callback(int opt, char const* value, void* data)
+cmdline_callback(int opt, char const* value, MD_UNUSED void* data)
```
rsms, line 297: I wasn't sure if this was a form of "documentation" so rather than removing this unused argument I decorated it so that the compiler knows its intentionally unused.

src/md4c.c
```diff
@@ -466,7 +472,7 @@ md_text_with_null_replacement(MD_CTX* ctx, MD_TEXTTYPE type, const CHAR* str, SZ
 typedef struct MD_UNICODE_FOLD_INFO_tag MD_UNICODE_FOLD_INFO;
 struct MD_UNICODE_FOLD_INFO_tag {
     unsigned codepoints[3];
-    int n_codepoints;
+    unsigned n_codepoints;
```
rsms, line 475: See https://github.com/mity/md4c/pull/132/commits/f9581a9a17e9eeb9ee0df1a2bf59594d48801b75

src/md4c.c
```diff
@@ -1232,7 +1238,8 @@ md_is_html_any(MD_CTX* ctx, const MD_LINE* lines, int n_lines, OFF beg, OFF max_
  ****************************/
 
 static int
-md_is_hex_entity_contents(MD_CTX* ctx, const CHAR* text, OFF beg, OFF max_end, OFF* p_end)
+md_is_hex_entity_contents(MD_UNUSED MD_CTX* ctx, const CHAR* text,
```
rsms, line 1241: the `ctx` argument is unused in several of these internal functions. Rather than removing that argument I thought it would be best to keep it as I'm not sure if there were personal opinions behind uniformity or documentation in this way. Personally I'd remove unused arguments but I leave that up to you @mity

src/md4c.c
```diff
@@ -5659,7 +5664,7 @@ md_line_indentation(MD_CTX* ctx, unsigned total_indent, OFF beg, OFF* p_end)
     return indent - total_indent;
 }
 
-static const MD_LINE_ANALYSIS md_dummy_blank_line = { MD_LINE_BLANK, 0 };
+static const MD_LINE_ANALYSIS md_dummy_blank_line = { MD_LINE_BLANK, 0, 0, 0, 0 };
```
rsms, line 5667: If you prefer implicit zero initialization I _think_ you could specify `-std=c11` (or `-std=c99` even.) AFAIK it was undefined behavior prior to C99. I might be wrong. Anyhow, simply initializing these seemed like the smallest and simplest fix.

mity:
Hello,

sorry for the late reply. I went finally over the patches, I am fine with some, not so with others:

> * 0812df7 - adds -watch option to scripts/build.sh for rebuilding as source files change
> * f4f8994 - adds -clean option to scripts/build.sh and places debug & release builds in separate subdirectories under build
> * 4e75722 - adds -test flag to scripts/build.sh to run tests after successful build
> * ec313b4 - adds CLI options to scripts/build.sh and strenghten compiler diagnostics
> * 8707b85 - adds scripts/build.sh for building the project

I use a similar script located in my `$PATH` and many other people do as well, so I am not that much in a favor in a project a script I never use/test/maintain at all.

> * af66338 - md2html: decorates data arg in cmdline_callback as unused
> * b78d49c - md4c: removes or declares unused function arguments

I can imagine marking the unused arguments, if it silences some compiler warnings. However, better to do it in a compiler-agnostic way:

```
void func(int unused_arg)
{
    (void) unused_arg;

    ...
}
```

So not merged.

> * 59efc10 - Adds missing field initializers (undefined behavior)

Merged.

> * f9581a9 - Fixes sign mismatch comparison

This change causes makes the test to report a regression:

```
Code coverage & regressions:
Example 17 (lines 186-191) [Issue 78](https://github.com/mity/md4c/issues/78)
[SS ẞ]: /url
[ẞ SS]
```

So not merged.

> * 31c91e2 - Initializes an uninitilized variable in md_analyze_emph

Merged, why not. Even though for any uninitialized variable there is always at least one compiler or its version or its combination of flags which warns about it being possibly used uninitialized...

> * a338a14 - adds /build dir to gitignore

Merged.

mity:
(Some commits cherry-picked inidividually. Please consider next time to create individual PRs for unrelated stuff.)

src/md4c.c
craigbarnes: FWIW, standard C has always had implicit zero initialization for braced initializers; even in C89. GCC complains because apparently it's considered "error prone" to not warn for missing elements, except for the special case of `{0}`.

PR #141, opened 2020-12-14
Build with -Wall -Wextra, and silence most warnings

dangelog:
As per subjet.

AMEND: actually, had to fix all the warnings, as the CI is quite strict in that regard (didn't spot that it was under -Werror).


mity:
> didn't spot that it was under `-Werror`

Yeah, a favor to Qt team who needs `-Wdeclaration-after-statement` to be treated as an error due some embedded compiler which chokes on that. If there is a better way how to turn just that one warning type into an error, I would prefer that.

> `n_codepoints` does not seem to ever contain negative offsets anyhow, should it actually be unsigned?

Strictly speaking yes. It was `int` for some historical reasons. I think -1 was used as a special value when end of string was reached or something like that. Looking at the code, it's not true anymore.

mity:
Ok, merged, seems ok to me. Thanks for working on it.

Issue #145, opened 2021-01-04
linking with the imported cmake targets does not work


niclasr:
When linking with cmake using the exported imported targets in a cmake file like this:

```
cmake_minimum_required(VERSION 3.18)
project(md4c-test LANGUAGES C)

find_package(md4c REQUIRED)
add_executable(md4ctest md4ctest.c)
target_link_libraries(md4ctest md4c)
```
the include directory is not added automatically to the compile command md4ctest. This can be fixed by adding
`INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}`
to the [INSTALL(TARGETS](https://cmake.org/cmake/help/v3.19/command/install.html#install-targets) command of the md4c and md4c-html targets.

Secondly md4cHtml or md4c-html package is not found at all. The reason for this is that the directory and the name of the config are different. Pick one such as md4c-html and have the config name changed to md4c-htmlConfig and it works or change the install directory to md4cHtml and it also works, they must be the same as can be seen in [the cmake docs on search order](https://cmake.org/cmake/help/v3.19/command/find_package.html#search-procedure). I suggest using the first since it is consistent with the target name.
I have made this an issue instead of a pull request since both ways work and I want to know which one you think is the best.

When the above works md4c html complains about not finding the target md4c that is wants to link to.  This can be fixed by adding PRIVATE to its target_link_libraries command but that might have implications on static linking, I guess.

I can make a pull request with the above changes if you are interested.

mity:
Oh yes, CMake. It's sometimes hard to make it work indepenently on the generator used or on the platform. And I'm the 1st to admit it's not my strong side: I use copy&paste there.

So yes, if you can cook a PR, it would be highly appreciated.

niclasr:
I have now made a PR #146 . It includes a fix for the last problem as well. I don't know what you think of the namespace addition since it might break backward compatibility.

PR #146, opened 2021-01-04
Fix use of the cmake package

niclasr:
Fix use of the cmake package and its imported targets.
Make sure that the include dir comes with the cmake targets
Put everything under md4cConfig so that the md4c-html can see
md4c.
Use md4c namespace so that the targets become md4c::md4c and
md4c::md4c-html following cmake standards for imported targets.

Since a prefix has been added to the targets maybe they should be
documented in the readme.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,7 +16,7 @@ set_target_properties(md4c PROPERTIES
     SOVERSION ${MD_VERSION_MAJOR}
     PUBLIC_HEADER md4c.h
 )
-
+add_library(md4c::md4c ALIAS md4c)
 
 # Build rules for HTML renderer library
 
@@ -27,7 +27,7 @@ set_target_properties(md4c-html PROPERTIES
     SOVERSION ${MD_VERSION_MAJOR}
     PUBLIC_HEADER md4c-html.h
 )
-target_link_libraries(md4c-html md4c)
+target_link_libraries(md4c-html md4c::md4c)
 
 
 # Install rules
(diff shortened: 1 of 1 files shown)


mity:
First of all, thank you for the work on it.

I never encountered the CMake namespaces yet, so forgive me asking few (potentially silly) questions.

* Actually, the fact alone I never encountered them so far makes me wonder how much their use is just a recommended practice by CMake authors from academic point of view; or how much widespread/standard it really is in a real-world use. Generally, I tend to be pragmatic and take the way of least surprise in such matters.

* Just to be sure, the change affects only CMake, more specifically how 3rd-party `CMakeLists.txt` files specify their targets to link against the lib and it affects "only" the build of CMake-based 3rd party applications, right? 

  I want to be especially certain there are no new library files built/installed and no changes in paths/names of the library binaries. MD4C is used already in [many linux distros](https://repology.org/project/md4c/versions) and update of it must not break a binary compatibility of already built/shipped applications.

* Why is there `add_library(md4c::md4c ALIAS md4c)` but no analogical line for `md4c-html`? Or put differently, where exactly is the `md4c-html` target made to live in the namespace?

> I don't know what you think of the namespace addition since it might break backward compatibility.

I'm C guy more than CMake guy and I never tried/needed to use CMake to its full potential. So right now I'm trying to understand how big the breakage really is (or whether there is any).

If I understand #145 correctly the `md4c-html` target has been completely broken and unusable, right? (I likely never needed to use that from another `CMakeLists.txt` except for `md2html` utility). If so, there cannot be any compatibility issue as it had to be broken from the very start when that lib was separated from `md2html` for better reuse.

And if the alias `md4c` for `md4c::md4c` allows to still refer to it with just `md4c`, there should be no compatibility issue introduced either, right?

niclasr:
> Just to be sure, the change affects only CMake, more specifically how 3rd-party CMakeLists.txt files specify their targets to link against the lib and it affects "only" the build of CMake-based 3rd party applications, right?

> I want to be especially certain there are no new library files built/installed and no changes in paths/names of the library binaries. MD4C is used already in many linux distros and update of it must not break the binary compatibility of already built/shipped applications.

Yes it only affects 3rdparty CMake builds that depend on md4c via `find_package(md4c)` command, no changes to the library names and the pkg-config files. 

> If I understand #145 correctly the md4c-html target has been completely broken and unusable, right? (I likely never needed to use that from another CMakeLists.txt except for md2html utility). If so, there cannot be any compatibility issue as it had to be broken from the very start when that lib was separated from md2html for better reuse.

Yes due to the difference between the directory name and the config name,  so moving the target md4c-html to md4cConfig is perfectly safe.

> Why is there add_library(md4c::md4c ALIAS md4c) but no analogical line for md4c-html? Or put differently, where exactly is the md4c-html target made to live in the namespace?

I thought this was needed for md4-html to be dependent on md4c::md4c in the generated md4cConfig. I just tested without it and made md4c-hml depend on md4c as usual, the generated file is the same so the line is useless, I will remove it.
The targets are made to live under the md4c:: namespace by the `NAMESPACE md4c::` in the install command on the last line.

> Actually, the fact alone I never encountered them so far makes me wonder how much their use is just a recommended practice by CMake authors from academic point of view; or how much widespread/standard it really is in a real-world use. Generally, I tend to be pragmatic and take the way of least surprise in such matters.

> And if the alias md4c for md4c::md4c allows to still refer to it with just md4c, there should be no compatibility issue introduced either, right?

The alias command used make md4c::md4c and alias form md4c and as mentioned above it is not needed.
Due to the namespace the exported target name will change and that breaks CMake builds that use the md4c target.
The Cmake authors write this about namespaces in [their export guide](https://cmake.org/cmake/help/v3.19/guide/importing-exporting/index.html?highlight=namespace#exporting-targets)
> The NAMESPACE option will prepend MathFunctions:: to the target names as they are written to the export file. This convention of double-colons gives CMake a hint that the name is an IMPORTED target when it is used by downstream projects. This way, CMake can issue a diagnostic message if the package providing it was not found.

Well this might not be enough reason for you to add a namespace that breaks backward compatibility with CMake builds that depends on md4c via `find_package(md4c)` .  If you want backward compatibility for 3rd party CMake builds I can just remove the `NAMESPACE md4c::` part and it works just like before, the choice is up to you.

mity:
> I just tested without it and made md4c-hml depend on md4c as usual, the generated file is the same so the line is useless, I will remove it.

Please do, and I'll merge.

> If you want backward compatibility for 3rd party CMake builds I can just remove the NAMESPACE md4c:: part and it works just like before, the choice is up to you.

If you think that (except the compatibility) it's better to use the namespace, lets keep it as it is, at least for now. If it happens people complain before the next release (probably not anytime soon) we might reconsider.

mity:
Merged, thank you for working on it.

PR #158, opened 2021-05-04
Allow user to customize generated HTML

strager:
md_html allows very little customization. If users want to use `<tt>`
instead of `<code>`, for example, they need to reimplement md_html
entirely or post-process md_html's output.

Expose some of md_html's internals to make it easier for users to
customize md_html's output.


mity:
Hello, 1st of all sorry for the very late response to this.

I find this as an interesting idea generally, but it needs some more work before I'm ready to merge this.

1. Most importantly, the new interface in `md4c-html.h` would have to be re-designed in some bigger way to allow accommodating future expansion of md4c.h interface. 

Imagine future version of the structure `MD_PARSER` gets a new (optional) member which allows using a new optional feature. The parser (md4c.c) would only use the new member and enable the feature only when `MD_PARSER::abi_version >= 1` (for compatibility with current applications out there.) Completely legitimate thing to do.

But such future feature then could never be used together with `md_html_create()` because we could not change interface of that function so it could not be able to initialize the new struct member(s).

2. Please avoid using C-99 and C-11 struct initializers in the code. In some cases MD4C is translated with very antique C compilers (for example Qt uses MD4C and they need to compile it for some embedded systems with obscure old C compiler).

3. I would also prefer to move md4c-html-example.c to a new dir (e.g. md4c/examples) to not clutter the main stuff. Example serves both partly a feature of source code and partly also that of a documentation so it would probably be better from that perspective too.

strager:
I don't plan to work on this patch. Feel free to take it over. 👍

PR #168, opened 2021-08-29
Fix buffer overflows and other errors found with fuzzying.

TCKnet:


mity:
Thanks.

(And sorry for the long reaction time.)

PR #171, opened 2021-10-08
Add mention links extension

niblo:
This adds support for mention links.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -3196,6 +3196,10 @@ md_collect_marks(MD_CTX* ctx, const MD_LINE* lines, int n_lines, int table_mode)
                 if( (ctx->parser.flags & MD_FLAG_MENTIONS) && (line->beg == off || (CH(off-1) == _T(' '))) )
                 {
                     OFF index = off + 1;
+                    if (index == line->end || CH(index) == ' ') {
+                        off++;
+                        continue;
+                    }
                     while (index <= line->end)
                     {
                         if (!(ISALNUM(index) || (CH(index) == '_')))
--- a/test/mention-links.txt
+++ b/test/mention-links.txt
@@ -15,3 +15,15 @@ someone said @me said such and such.
 .
 <p><x-mention data-target="me">@me</x-mention> said such and such.</p>
 ````````````````````````````````
+
+```````````````````````````````` example
+An empty at character is not recognized as a mention:
+
+@ 
+
+@
+.
+<p>An empty at character is not recognized as a mention:</p>
+<p>@</p>
+<p>@</p>
+````````````````````````````````


niblo:
I'm no longer getting a segmentation fault. It was something downstream.

mity:
Just tried it. Seems to crash more or less with any input containing `@` now (when the extension is enabled).

Here for the input `@x`:

```
(gdb) bt
#0  0x0040d0f9 in md_process_inlines (ctx=0x64faa8, lines=0x671510, n_lines=1) at ../src/md4c.c:4318
#1  0x0040e2a8 in md_process_normal_block_contents (ctx=0x64faa8, lines=0x671510, n_lines=1) at ../src/md4c.c:4639
#2  0x0040ea42 in md_process_leaf_block (ctx=0x64faa8, block=0x671508) at ../src/md4c.c:4818
#3  0x0040ed86 in md_process_all_blocks (ctx=0x64faa8) at ../src/md4c.c:4900
#4  0x0041246e in md_process_doc (ctx=0x64faa8) at ../src/md4c.c:6333
#5  0x0041262d in md_parse (
    text=0x672dc8 "@x­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş"..., size=2, parser=0x64fce4, userdata=0x64fd08) at ../src/md4c.c:6401
#6  0x00403cef in md_html (
    input=0x672dc8 "@x­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş"..., input_size=2, process_output=0x401e59 <process_output>, userdata=0x64fe60,
    parser_flags=32768, renderer_flags=5) at ../src/md4c-html.c:580
#7  0x00401f87 in process_file (in=0x760c4660 <msvcrt!_iob+96>, out=0x760c4620 <msvcrt!_iob+32>)
    at ../md2html/md2html.c:144
#8  0x004026fe in main (argc=3, argv=0x672c78) at ../md2html/md2html.c:380
```

src/md4c-html.c
```diff
@@ -265,6 +265,13 @@ render_attribute(MD_HTML* r, const MD_ATTRIBUTE* attr,
     }
 }
 
+static void
+render_mention_link(MD_HTML* r, const MD_SPAN_MENTION_DETAIL* det)
+{
+    RENDER_VERBATIM(r, "<x-mention data-target=\"");
+    render_entity(r, det->text, det->size, render_html_escaped);
```
mity, line 272: Should likely call `render_verbatim()` instead. No entity translation should place here I guess.

src/md4c.c
```diff
@@ -3190,9 +3190,26 @@ md_collect_marks(MD_CTX* ctx, const MD_LINE* lines, int n_lines, int table_mode)
...
+                    while (index <= line->end)
+                    {
+                        if (!(ISALNUM(index) || (CH(index) == '_')))
+                            break;
+                        index++;
+                    }
+                    PUSH_MARK('@', off, index, MD_MARK_RESOLVED);
+                    off = index;
+                }
+                else if(line->beg + 1 <= off  &&  ISALNUM(off-1)  &&
```
mity, line 3212: This condition now should likely explicitly check whether are enabled `MD_FLAG_PERMISSIVEEMAILAUTOLINKS`. That was not needed previously because `@` was added into the `mark_char_map[]` in `md_build_mark_char_map()` if and only if that extension was enabled.

src/md4c.c
```diff
@@ -4299,6 +4317,17 @@ md_process_inlines(MD_CTX* ctx, const MD_LINE* lines, int n_lines)
...
+                    if (CH(mark->beg) == '@')
+                    {
+                        det.text = (char *) ctx->text + mark->beg + 1;
+                        det.size = mark->end - mark->beg - 1;
+                        MD_ENTER_SPAN(MD_SPAN_MENTION, &det);
+                        MD_TEXT(text_type, STR(mark->beg), mark->end - mark->beg);
+                        MD_LEAVE_SPAN(MD_SPAN_MENTION, &det);
+                        break;
+                    }
+
```
mity, line 4330: How do you know at this point it is a mention link and a permissive e-mail auto-link? Both extensions may be enabled at the same time.

src/md4c.h
```diff
@@ -145,7 +145,9 @@ typedef enum MD_SPANTYPE {
 
     /* <u>...</u>
      * Note: Recognized only when MD_FLAG_UNDERLINE is enabled. */
-    MD_SPAN_U
+    MD_SPAN_U,
+
+    MD_SPAN_MENTION
```
mity, line 150: I would prefer renamimg it to `MD_SPAN_MENTIONLINK` for the sake of consistency e.g. with `MD_SPAN_WIKILINK` etc.

src/md4c.h
```diff
@@ -297,6 +299,12 @@ typedef struct MD_SPAN_WIKILINK {
     MD_ATTRIBUTE target;
 } MD_SPAN_WIKILINK_DETAIL;
 
+/* Detailed info for MD_SPAN_MENTION. */
+typedef struct MD_SPAN_MENTION {
+    unsigned char size; 
+    MD_CHAR* text;
```
mity, line 305: I understand it's supposed to contain only a verbatim text (the username) with no possible nested formatting options, yet I wonder whether rather using `MD_ATTRIBUTE target` (and initializing it to a single `substr_type` of type `MD_TEXT_NORMAL`) would not be better, also for the sake of consistency with other detail structures.

src/md4c.h
```diff
@@ -316,6 +324,7 @@ typedef struct MD_SPAN_WIKILINK {
 #define MD_FLAG_LATEXMATHSPANS              0x1000  /* Enable $ and $$ containing LaTeX equations. */
 #define MD_FLAG_WIKILINKS                   0x2000  /* Enable wiki links extension. */
 #define MD_FLAG_UNDERLINE                   0x4000  /* Enable underline extension (and disables '_' for normal emphasis). */
+#define MD_FLAG_MENTIONS                    0x8000  /* Enable mention links extension. */
```
mity, line 327: Similarly please rename to `MD_FLAG_MENTIONLINKS`.

mity (changes requested):
FYI, this is not a full review, take it as just a few cents of what I've just noticed when skimming through the code quickly. I likely won't do anything deeper until you fix the crashing so I can try it more thoroughly to assess it more.

mity (changes requested):
Noticed one more thing.

niblo:
> Just tried it. Seems to crash more or less with any input containing `@` now (when the extension is enabled).
> 
> Here for the input `@x`:
> 
> ```
> (gdb) bt
> #0  0x0040d0f9 in md_process_inlines (ctx=0x64faa8, lines=0x671510, n_lines=1) at ../src/md4c.c:4318
> #1  0x0040e2a8 in md_process_normal_block_contents (ctx=0x64faa8, lines=0x671510, n_lines=1) at ../src/md4c.c:4639
> #2  0x0040ea42 in md_process_leaf_block (ctx=0x64faa8, block=0x671508) at ../src/md4c.c:4818
> #3  0x0040ed86 in md_process_all_blocks (ctx=0x64faa8) at ../src/md4c.c:4900
> #4  0x0041246e in md_process_doc (ctx=0x64faa8) at ../src/md4c.c:6333
> #5  0x0041262d in md_parse (
>     text=0x672dc8 "@x­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş"..., size=2, parser=0x64fce4, userdata=0x64fd08) at ../src/md4c.c:6401
> #6  0x00403cef in md_html (
>     input=0x672dc8 "@x­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş\rđ­ş"..., input_size=2, process_output=0x401e59 <process_output>, userdata=0x64fe60,
>     parser_flags=32768, renderer_flags=5) at ../src/md4c-html.c:580
> #7  0x00401f87 in process_file (in=0x760c4660 <msvcrt!_iob+96>, out=0x760c4620 <msvcrt!_iob+32>)
>     at ../md2html/md2html.c:144
> #8  0x004026fe in main (argc=3, argv=0x672c78) at ../md2html/md2html.c:380
> ```

Rebased on the latest master, I'm not getting any errors when running the tests. This is on OS X. I see you're running on Windows. Do any of the checks (Travis CI, codecov, ...) run the test suite?

mity:
> Rebased on the latest master, I'm not getting any errors when running the tests. This is on OS X. I see you're running on Windows. 

Still crashing on Windows. 

But I can see that if I move the block of lines 4315 - 4318 after that following mention-related `if`, the crash goes away. That opener/closer arithmetics likely makes no sense in the mention case (it has no real closer, right?). 

I have quite old gcc version on windows, so I can imagine newer compiler versions reorder that code as an optimization because the mention branch block does not depend on those variables and thus hide the bug effectively.

> Do any of the checks (Travis CI, codecov, ...) run the test suite?

At this time, Travis only (linux) does the tests. Not sure how easy it would be to enable them on AppVeyor (windows) as it requires python but I hope it should be possible.

mity:
Update: On Linux the problem is there too, only it does not manifest by the crash for some reason. Whenever running it under valgrind, enabling mentions generates invalid reads from unallocated memory.

niblo:
> 

Which lines exactly?

mity:
> Which lines exactly?

These 4 lines make no sense to me in the case of input made of only mentions. That pointer arithmetic leads to the problem.  There must be some logic added makeing that code work for links and auto-links but not to be executed for the case of mention which seem to have no closer (or alternatively you need to change the logic that some virtual empty closer marks are created even for the mentions).

https://github.com/niblo/md4c/blob/3f355de536520791c844c2fa95041b897ae7bc44/src/md4c.c#L4315-L4318

niblo:
> > Which lines exactly?
> 
> These 4 lines make no sense to me in the case of input made of only mentions. That pointer arithmetic leads to the problem. There must be some logic added makeing that code work for links and auto-links but not to be executed for the case of mention which seem to have no closer (or alternatively you need to change the logic that some virtual empty closer marks are created even for the mentions).
> 
> https://github.com/niblo/md4c/blob/3f355de536520791c844c2fa95041b897ae7bc44/src/md4c.c#L4315-L4318

I'm having some issues compiling on Windows. How do you do it?

mity:
> I'm having some issues compiling on Windows. How do you do it?

I'm generally using mingw-w64, or more specifically very old MSYS (not MSYS2) environment with gcc-toolchain built from https://github.com/niXman/mingw-builds/ and I'm build with Cmake+Ninja. Unfortunately that can be hard-to-replicate environment.

That said though, you should be able to build with MSYS 2 (can be downloaded from https://mingw-w64.org) or directly with any recent version of MS Visual Studio (afaik, it now supports building CMake-based projects directly). Or you can generate MSVC solution manually by CMake as appveyor.xml does for the purposes of CI.

However note I don't think that making some cosmetic change preventing the most obvious crashes is good enough approach here. The pointed piece of code shows this PR makes a fundamental difference between how the mentions are implemented and how other permissive links are. 

For the permissive e-mail auto-links we add an extra dummy mark for every potential valid `@` (when handling `@` in `md_collect_marks()`), and in case we decide later to resolve it as an actual link we turn that dummy mark into a virtual (zero-length) closer mark. That means all the later link processing works quite independently on whether it was a full markdown link or permissive auto-link because they are encoded similarly in the `MD_CTX::marks[]`.

I don't think it's a good idea to depart from this approach for mention links. It could lead to difficult-to-solve collisions of such radically different approaches in all phases of the processing. So please take a look whether mentions could work the same way: In ideal case you would then need to distinguish only at the last moment what constants to use when calling the callback functions for one or the other. I think it should be possible.

ec1oud:
x-mention-data-target is an html tag already in use?

mity:
> x-mention-data-target is an html tag already in use?

No. Only in this PR. Even even when/if it gets merged you would see it only when enabled with an extension flag.

PR #182, opened 2022-08-23
Defensively fix 3 asserts that didn't hold and prevent endless loop in md_rollback

TCKnet:
These were discovered through fuzzying


TCKnet:
Test cases: [testcases.zip](https://github.com/mity/md4c/files/9406790/testcases.zip)

PR #185, opened 2022-11-24
Update md4c.c

TCKnet:
Fix a heap-buffer-overflow on the following data (fuzzed).

const char data[] = { 127, 247, 0, 0, 0, 4, 191, 180, 89,
  10, 45, 124, 10, 124, 120, 64, 101, 46, 101, 36, 93, 36 };


TCKnet:
Here is the fuzzer data to replay the two crashes, the timeout (last test / does not reproduce on oss-fuzz) is caused by a huge chain of dependencies in md_analyze_marks (fixed by the capping at 2000 at line 4009).

 const char data[] = {
39, 35, 10, 51, 44, 122, 61, 124, 124, 127, 92, 201, 91, 91, 33, 91, 85, 95, 42, 95, 42, 95, 95, 58, 95, 58, 95, 95, 58, 95, 42, 95, 58, 95, 58, 55, 53, 95, 95, 58, 95, 42, 95, 58, 95, 95, 42, 95, 58, 95, 58, 95, 58, 42, 95, 42, 95, 95, 58, 95, 58, 95, 95, 58, 95, 183, 95, 58, 95, 42, 95, 42, 124, 93, 40, 41, 93, 93, 201, 91, 91, 33, 91, 97, 97, 41, 49, 97, 47, 60, 85, 95, 42, 95, 42, 95, 95, 58, 95, 58, 95, 95, 58, 95, 42, 95, 58, 95, 58, 95, 58, 42, 95, 42, 95, 97, 95, 58, 95, 42, 95, 58, 95, 95, 42, 58, 95, 58, 42, 95, 42, 95, 58, 95, 58, 116, 95, 102, 95, 58, 95, 42, 95, 58, 95, 42, 95, 42, 124, 93, 40, 41, 93, 93, 10, 45, 45, 45, 124, };
 const char data[] = {
181, 181, 42, 0, 10, 60, 62, 10, 124, 91, 42, 95, 42, 93, 40, 41, 10, 45, 124, };

mity:
I've failed to reproduce any of those.

Also the PR here is not really a fix: Those checked conditions would indicate broken internal state, i.e. real fix would be to make sure MD4C does not get into that state.

mity:
Ahh, I see: These are the cases from the oss-fuzz so the initial bytes specify parser and renderer options. Hopefully should be fixed now. 

Closing this PR as the suggested changes don't really address the root causes but make random checks for broken internal state which should never really happen.

Issue #192, opened 2023-02-25
Project dead?


tim-gromeyer:
I'm following the project for some time now and the creator hasn't commited for over a years now neighter has he wrote any comments on issues/pull requests.

That's why I'm windering if this project is dead.

dominickpastore:
I hope not. It didn't seem to be a slow fizzling out, either&mdash;he was pretty active, until he wasn't. Hope he's doing alright.

tim-gromeyer:
He seems pretty active on his [twitter account](https://twitter.com/MartinMitas/with_replies). Maybe we can contact him there. I can't, I don't have a Twitter account.

dominickpastore:
I sent him a DM there but didn't get a response :/

vtorri:
he seems quite involved in war comments. @dominickpastore can you try again ?

acidvegas:
Bumping for interest

tim-gromeyer:
Has someone plans to create a fork if it?

tim-gromeyer:
I have a [fork](https://github.com/tim-gromeyer/MarkdownEdit_md4c) with #175 for my [Markdown editor MarkdownEdit](https://github.com/tim-gromeyer/MarkdownEdit) but I'm not actively developing it

PR #193, opened 2023-06-07
Add option `MD_FLAG_HARD_SOFT_BREAKS`

l1mey112:
### [Soft Line Breaks - CommonMark Spec](https://spec.commonmark.org/0.30/#soft-line-breaks)

The ability to tread soft line breaks as hard line breaks is something specified in the spec, and highly needed for my project. Thank you for creating this library.

### DIFF
--- a/test/hard-soft-breaks.txt
+++ b/test/hard-soft-breaks.txt
@@ -1,0 +1,23 @@
+
+# Hard Soft Breaks
+
+With the flag `MD_FLAG_HARD_SOFT_BREAKS`, MD4C treats all newline characters as soft breaks.
+
+```````````````````````````````` example
+foo
+baz
+.
+<p>foo<br>
+baz</p>
+````````````````````````````````
+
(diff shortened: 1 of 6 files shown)


l1mey112:
@mity It seems like there hasn't been any commits since early 2022, is there any reason for that?

tim-gromeyer:
@l1mey112 it seems like this project is dead, unfortunately :disappointed: 

You can check out #192

test/hard-soft-breaks.txt
```diff
@@ -0,0 +1,23 @@
+
+# Hard Soft Breaks
+
+With the flag `MD_FLAG_HARD_SOFT_BREAKS`, MD4C treats all newline characters as soft breaks.
```
step-, line 4: Shouldn't it read "...as hard breaks" ?

mity: It should, thx.

Issue #202, opened 2024-01-08
Leave callback not called for HTML block comment


step-:
For the specific case of a one-line block comment (HTML block type 2) not followed by a blank line[1], MD4C fails to call the leave callback. Consequently, if another HTML comment follows the first comment, both end up being combined into a single block instead of staying as two separate blocks, each with its own callback.

[1] In my comments to issue https://github.com/mity/md4c/issues/200 I remarked that the spec doesn't require a blank line to close a
type 2 HTML block.

DETAILS

I can't think of a simple way to demonstrate this issue using md2html alone, precisely because the lack of something can't be shown. Feeding md2html the following markdown:

```
<!-- C1 -->
<!-- C2 -->
```

outputs the input text so one would be inclined to think that everything is correct. However, what can't be seen is that there is no callback between lines C1 and C2, while there should be one. Instead, a callback after line C2 wrongly combines the two single-line blocks into a two-line block. Trace the code or add printf statements as needed to see the issue at work.

mity:
Thanks for the report. If I understand correctly, the issue is that the implementation sort of merges two HTML blocks together if they follow one after another, and calls the `enter_block` just once before the 1st one, and `leave_block` after the last one.

However I wonder what your expectations really are and whether we need to fix this at all. Consider especially that [the type 7 of HTML blocks](https://spec.commonmark.org/0.30/#html-blocks) may contain not one but whole set of HTML tags; and it may even be an invalid HTML chunk after the initial tag condition. All what matters is whether it satisfies the opening and closing conditions.

I.e. what the `enter_block(MD_BLOCK_HTML)` tells the renderer is: Whatever you get in the `text_callback` should be treated as a raw HTML, until you see `leave_block(MD_BLOCK_HTML)`. However it does not generally break the HTML into individual tags and does not even attempt any HTML validation of it beyond the Markdown specification, and it wouldn't (at least for the type 7) even after "fixing" this.

step-:
You understood correctly but enlarged the scope to all HTML blocks. I'm only concerned about type 2 HTML blocks. Not interested in type 7. Indeed PR #203 specifically uses the value of `html_block_type` (2) to restrict the scope of my change.

step-:
My expectation is that it gets fixed for type 2 HTML blocks because cmark renders every HTML block comment as a separate block, and the application I'm developing on top of MD4C relies on the same behavior.

Issue #200, opened 2023-12-09
Corrupted code block after HTML block comment


step-:
Example https://spec.commonmark.org/0.30/#example-134
Running md2html with just the example and no other text included meets the spec. However, when I add an HTML block before the example the test result changes unexpectedly and fails the spec.
Failing test subject:
```````````````````````````````` subject
<!-- x -->
    ```
    aaa
    ```
````````````````````````````````
Expected output:
```````````````````````````````` reference
<!-- x -->
<pre><code>```
aaa
```
</code></pre>
````````````````````````````````

Actual output (fail):
```````````````````````````````` result
<!-- x -->
<pre><code>aaa
```
</code></pre>
````````````````````````````````

----

Here's another example of the same problem. This time the HTML comment spans two lines. As in the first example, md2html is missing some text in the rendered code block.

```````````````````````````````` subject
<!-- x
-->
    ```
    aaa
    ```
````````````````````````````````
Expected output:
```````````````````````````````` reference
<!-- x
-->
<pre><code>```
aaa
```
</code></pre>
````````````````````````````````

Actual output (fail):
```````````````````````````````` result
<!-- x
-->
<pre><code>aaa
```
</code></pre>
````````````````````````````````

N-R-K:
Indented code blocks need to be separated by black lines: https://spec.commonmark.org/0.30/#indented-code-blocks

step-:
> Indented code blocks need to be separated by black lines: https://spec.commonmark.org/0.30/#indented-code-blocks

False. I read that section differently. It says that there must be a blank line before the block to interrupt a preceding _paragraph._  The failing example concerns a type 2 _HTML block_, https://spec.commonmark.org/0.30/#html-blocks, before an indented code block. In the failing example, the HTML block is properly ended. Then comes an indented code block, which is rendered incorrectly. Run the failing example under `cmark` yields the reference output shown in my report.

step-:
As a follow-up to @N-R-K's [comment](https://github.com/mity/md4c/issues/200#issuecomment-1848800853) I verified md2html vs [cmark](https://github.com/commonmark/cmark) for the following failing test subject:

`````
start
    ```
    aaa
    ```
`````

[...]

PR #203, opened 2024-01-08
Fix leave callback not called for HTML block comment.

step-:
For the specific case of a one-line block comment (HTML block type 2) not followed by a blank line[1], MD4C fails to call the leave callback. Consequently, if another HTML comment follows the first comment, both end up being combined into a single block instead of staying as two separate blocks, each with its own callback.

[1] In my comments to issue https://github.com/mity/md4c/issues/200 I remarked that the spec doesn't require a blank line to close a type 2 HTML block.

DETAILS

I can't think of a simple way to demonstrate this issue using md2html alone, precisely because the lack of something can't be shown. Feeding md2html the following markdown:

```
<!-- C1 -->
<!-- C2 -->
```

outputs the input text so one would be inclined to think that everything is correct. However, what can't be seen is that there is no callback between lines C1 and C2, while there should be one. Instead, a callback after line C2 wrongly combines the two single-line blocks into a two-line block. Trace the code or add printf statements as needed to see the issue at work.

This commit fixes the problem by setting `ctx->html_block_type` to a negative value as a special way to flag this end-of-block case.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -6080,7 +6080,7 @@ md_analyze_line(MD_CTX* ctx, OFF beg, OFF* p_end,
                 /* The line itself also may immediately close the block. */
                 if(md_is_html_block_end_condition(ctx, off, &off) == ctx->html_block_type) {
                     /* Make sure this is the last line of the block. */
-                    ctx->html_block_type = 0;
+                    ctx->html_block_type *= -1;
                 }
 
                 line->type = MD_LINE_HTML;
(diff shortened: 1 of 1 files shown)


step-:
Fixes #202.

mity:
I'm afraid, this PR goes in bad direction because it wrongly assumes a HTML comment never occupies more than one line.

Consider e.g. this:
```
<!--
comment 1
-->
<!--
comment 2
-->
```

I should commit a proper fix shortly.

step-:
It doesn't. The value of `html_block_type` is negated only in the statement that concerns an HTML block comment that starts and ends on the same line.

mity:
> It doesn't. The value of html_block_type is negated only in the statement that concerns an HTML block comment that starts and ends on the same line.

Proper fix is not to merge multiple blocks no matter whether they occupy one or multiple lines.

step-:
As I wrote before, the change in this PR can't apply to multi-line comments. Multi-line comments are treated elsewhere in your code. For single line comments, I believe this fix is proper.

mity:
> Perhaps you mean you know that two consecutive multi-line comments miss the intermediate callback? That I don't know but I can easily verify...

That's what I mean. 319631f67e3ab44ec590bf66e424436a5f8dff1e works in that case too, and imho is also less hacky.

step-:
I understand you know. Meanwhile, I verified that the callback is missing between multi-line comments. I welcome the change.

step-:
Yes, I tested 319631f with my test suite and it works. Thank you.

PR #204, opened 2024-01-08
Fix corrupted code block after HTML comment

step-:
Close https://github.com/mity/md4c/issues/200.


mity:
Also fixed few minutes ago myself.

Issue #201, opened 2024-01-08
Code span wrongly identified as code block


step-:
The following test subject:

`````
start
    ```
    aaa
    ```
`````

is incorrectly rendered as a code block after a paragraph (mentioned in https://github.com/mity/md4c/issues/200#issuecomment-1880781719).

According to the spec, the whole subject constitutes a single paragraph that starts with "start" and ends with the second "```".
Below we can see that [cmark](https://github.com/commonmark/cmark) renders the test subject as a single paragraph, stripping white space before lines 2, 3 and 4, which then render as a code span. This is the correct way.

cmark:
`````
<p>p
<code>aaa</code></p>
`````

Below we can see that md2html renders the test subject incorrectly as a code block, and misses some original text, too.

md2html:
`````
<p>p</p>
<pre><code>aaa
```
</code></pre>
`````

PR #205, opened 2024-01-08
Fix code span wrongly identified as code block

step-:
Close https://github.com/mity/md4c/issues/201.

What happens in #201 is that MD4C identifies the first "```" as the start of a fenced code block.

Threfore it renders "aaa" and "    ```" as code block lines, then it
automatically closes the fenced code block at EOF.

DETAILS

Commonmark spec 0.30 about fenced code blocks.

A fenced code block begins with a code fence, preceded by up to three spaces of indentation.[1]
A fenced code block may interrupt a paragraph, and does not require a blank line either before or after.[2]

- [1] https://github.com/commonmark/commonmark-spec/blob/d5706b9553d4665ca730524323d484880440186c/spec.txt#L1940
- [2] https://github.com/commonmark/commonmark-spec/blob/d5706b9553d4665ca730524323d484880440186c/spec.txt#L1969

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -6058,7 +6058,7 @@ md_analyze_line(MD_CTX* ctx, OFF beg, OFF* p_end,
         }
 
         /* Check whether we are starting code fence. */
-        if(off < ctx->size  &&  ISANYOF2(off, _T('`'), _T('~'))) {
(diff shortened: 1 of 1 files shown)


mity:
Race condition :-)  Fixed few minutes ago myself.

PR #232, opened 2024-01-26
Use ptrdiff_t for pointer-difference types

ec1oud:
This fixes warnings on msvc in several places:

md4c.c(3087): warning C4244: 'function': conversion from '__int64' to 'int', possible loss of data

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -26,6 +26,7 @@
 #include "md4c.h"
 
 #include <limits.h>
+#include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


ec1oud:
I hope it's portable enough.  https://stackoverflow.com/questions/3597743/where-is-ptrdiff-t-defined-in-c

mity:
Thanks for it.

However I have to admit I dislike the patch because it now uses `ptrdiff_t` on numerous places which have nothing to do with pointer difference and I find the result misleading and not really improving quality of the code.

I've created an alternative patch based on this, if youre ok with it. See PR #233.

mity:
Closing, as the #233 was merged instead.

PR #255, opened 2024-05-24
Add CMake BUILD_MD4C_HTML whether to build md4c-html

mrjoel:
The README instructions provide guidance on optionally enabling or copying source for  md4c-html if needed. This adds a CMake option (defaulted to on) to enabling controlling whether the html lib is built.


mity:
Closing, the PR #392 implements this (in another way, to handle the options more consistently in the project and to-handle inter-dependency of the optional targets).

PR #265, opened 2024-08-21
Fix md_decode_utf16le_before__() return value

### LINKED ISSUE #275: `md_decode_utf16le_before__()` returns wrong value
I believe `md_decode_utf16le_before__()` returns the wrong value. The function is supposed to return the character before the offset. We can compare with the utf8 version [`md_decode_utf8_before__()`](https://github.com/mity/md4c/blob/481fbfbdf72daab2912380d62bb5f2187d438408/src/md4c.c#L921) which returns `CH(off-1)`.

I have created a PR here. https://github.com/mity/md4c/pull/265

MakotoE:
I believe `md_decode_utf16le_before__()` returns the wrong value. The function is supposed to return the character before the offset. We can compare with the utf8 version [`md_decode_utf8_before__()`](https://github.com/mity/md4c/blob/481fbfbdf72daab2912380d62bb5f2187d438408/src/md4c.c#L921) which returns `CH(off-1)`.

Closes https://github.com/mity/md4c/issues/275

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -855,7 +855,7 @@ struct MD_UNICODE_FOLD_INFO_tag {
         if(off > 2 && IS_UTF16_SURROGATE_HI(CH(off-2)) && IS_UTF16_SURROGATE_LO(CH(off-1)))
             return UTF16_DECODE_SURROGATE(CH(off-2), CH(off-1));
 
-        return CH(off);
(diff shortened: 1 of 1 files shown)


s77rt:
Test case that this PR fixes: `**text**abc`
Expected results: `text` to be in bold

PR #290, opened 2025-07-20
Fix alignment issue in md_mark_store_ptr()

### LINKED ISSUE #289: Alignment issue detected by Fil-C
When running under [Fil-C](https://github.com/pizlonator/llvm-project-deluge/blob/deluge/Manifesto.md) I'm getting
```
filc safety error: alignment requirement of 8 bytes not met.
    pointer: 0x7f65e3da5444,0x7f65e3da53e0,0x7f65e3da58e0,aux=0x7f65e3da5a20
    expected 17 writable bytes with ptr aligned to 8 bytes.
semantic origin:
    ../thirdparty/md4c/src/md4c.c:2648:5: md_mark_store_ptr (inlined)
    ../thirdparty/md4c/src/md4c.c:3623:13: md_resolve_links (inlined)
    ../thirdparty/md4c/src/md4c.c:4122:5: md_analyze_inlines
semantic origin:
    ../thirdparty/md4c/src/md4c.c:3622:13: md_resolve_links (inlined)
    ../thirdparty/md4c/src/md4c.c:4122:5: md_analyze_inlines
check scheduled at:
    ../thirdparty/md4c/src/md4c.c:3622:13: md_resolve_links (inlined)
    ../thirdparty/md4c/src/md4c.c:4122:5: md_analyze_inlines
    ../thirdparty/md4c/src/md4c.c:4699:5: md_process_normal_block_contents
    ../thirdparty/md4c/src/md4c.c:4878:13: md_process_leaf_block (inlined)
    ../thirdparty/md4c/src/md4c.c:4960:13: md_process_all_blocks (inlined)
    ../thirdparty/md4c/src/md4c.c:6414:5: md_process_doc (inlined)
    ../thirdparty/md4c/src/md4c.c:6483:11: md_parse
    ../src/markdown.c:580:9: convert_markdown_to_html
    ../src/blogpost.c:31:16: blogpost_html_into (inlined)
    ../src/blogpost.c:80:5: blogpost_new
    ../src/config.c:131:33: handle_blogpost
    ../src/filesystem_helpers.c:74:13: for_each_item_in_dir
    ../src/config.c:88:17: handle_config_directory_list
    ../src/config.c:177:5: config_load
    ../src/main.c:10:22: main
    src/env/__libc_start_main.c:79:7: __libc_start_main
    <runtime>: start_program
```

This seems to be due to
```c
/* Sometimes, we need to store a pointer into the mark. It is quite rare
 * so we do not bother to make MD_MARK use union, and it can only happen
 * for dummy marks. */
static inline void
md_mark_store_ptr(MD_CTX* ctx, int mark_index, void* ptr)
{
    MD_MARK* mark = &ctx->marks[mark_index];
    MD_ASSERT(mark->ch == 'D');

    /* Check only members beg and end are misused for this. */
    MD_ASSERT(sizeof(void*) <= 2 * sizeof(OFF));
    memcpy(mark, &ptr, sizeof(void*));
}
```

wmww:
Fixes #289 by making `MD_MARK_tag` a union (so the compiler will give it the correct alignment). Since the main member of the union is an anonymous struct, its properties can still be accessed as if they're toplevel `MD_MARK_tag` properties.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -147,7 +147,7 @@
 #define SZ_MAX      (sizeof(SZ) == 8 ? UINT64_MAX : UINT32_MAX)
 #define OFF_MAX     (sizeof(OFF) == 8 ? UINT64_MAX : UINT32_MAX)
 
-typedef union MD_MARK_tag MD_MARK;
+typedef struct MD_MARK_tag MD_MARK;
 typedef struct MD_BLOCK_tag MD_BLOCK;
 typedef struct MD_CONTAINER_tag MD_CONTAINER;
 typedef struct MD_REF_DEF_tag MD_REF_DEF;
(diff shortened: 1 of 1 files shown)


N-R-K:
Both the store and load are done via `memcpy` - which works at byte level - and thus there's no alignment issues. Whatever tool you're using is giving false positive warning.

wmww:
I would expect it to be well defined, I suppose a false positive is a possibility. In any case I think my patch (especially now after edits) makes the code clearer, simpler and shorter as well as making it play nice with Fil-C, so maybe still worth merging.

PR #296, opened 2026-01-28
Avoid repeated language- in info string.

### LINKED ISSUE #295: Avoid repeated language- in info string.
Do not add the language- prefix when the info string already starts with language-

See [commonmark/commonmark.js#277](https://github.com/commonmark/commonmark.js/issues/277)

znewt99:
Avoid repeated language- in info string.
Fix #295
See [commonmark/commonmark.js#277](https://github.com/commonmark/commonmark.js/issues/277).


mity:
Technically it's against the spec, but makes so much sense that I'm merging anyway.

Thx for it.

PR #297, opened 2026-03-11
Fix incorrect heading interpretation in math spans

rafal-c:
Prevent Setext heading interpretation within LaTeX math spans. Add test cases for such scenarios.

It's common to write LaTeX equations like this:
```
$$
lhs
=
rhs
$$
```
in master `$$ lhs` would be formatted as a heading, which is clearly incorrect.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -5267,6 +5267,43 @@ md_is_setext_underline(MD_CTX* ctx, OFF beg, OFF* p_end, unsigned* p_level)
 {
     OFF off = beg + 1;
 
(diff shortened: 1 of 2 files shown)


mity:
Hmmm. Sorry I won't  merge this. It's quite crazy code. And more importantly, it goes absolutely against CommonMark specification's [precendence principle](https://spec.commonmark.org/0.31.2/#precedence).

I'm not sure how to address the issue at the moment, but this is not the right way.

Also note GFM here also sees it as a title.

PR #312, opened 2026-05-04
Fix O(n²) performance regression in spoiler span analysis

### LINKED ISSUE #311: Pathologic input leading to heavily sub-linear parsing times
[OSS-fuzz](https:/github.com/google/oss-fuzz) found the following 661kB testcase, which on my machine takes 13+ seconds to parse: [clusterfuzz-testcase-fuzz-mdhtml-5579634954797056.zip](https://github.com/user-attachments/files/27372682/clusterfuzz-testcase-fuzz-mdhtml-5579634954797056.zip).

It can be reproduced by the semi-secret option `md2html --replay-fuzz <filename>`.

It seems to be regression caused by the new spoilers extension (so the recent 0.5.3 release is fine).

cc @hryhoriiK97

hryhoriiK97:
`md_analyze_link_contents()` is called recursively for every resolved link
pair. Each call runs the spoiler loop over its mark range, but unlike
`md_analyze_marks()`, the loop had no skip optimization for resolved spans
— it walked through every inner mark one by one.

With 64,135 resolved links in the OSS-Fuzz testcase, this caused
4.86 billion loop iterations instead of 268,499 — O(n²) behavior.

**Fix:** when the loop hits a resolved `[` or `!` opener, jump to
`mark->next` directly, mirroring what `md_analyze_marks()` already does

Fixes: #311

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -4173,8 +4173,15 @@ md_analyze_link_contents(MD_CTX* ctx, const MD_LINE* lines, MD_SIZE n_lines,
     if(ctx->parser.flags & MD_FLAG_SPOILERS) {
         for(i = mark_beg; i < mark_end; i++) {
             MD_MARK* mark = &ctx->marks[i];
-            if(mark->flags & MD_MARK_RESOLVED)
+            if(mark->flags & MD_MARK_RESOLVED) {
+                /* Skip resolved link/image spans. Their contents were already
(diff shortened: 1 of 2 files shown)


mity:
Yep, great job. Thx.

I've somewhat refactored `md_analyze_link_contents()` to reuse `md_analyze_marks()` instead of duplicating its logic. It lead also to some other code cleanup.

PR #313, opened 2026-05-05
Fix python SyntaxWarning in test

qaqland:
```
/home/runner/work/md4c/md4c/test/normalize.py:21: SyntaxWarning: invalid escape sequence '\s'
  whitespace_re = re.compile('\s+')
```

Refs: https://docs.python.org/3/whatsnew/3.12.html#other-language-changes


mity:
Thanks.

PR #314, opened 2026-05-07
Small Changes

znewt99:
Move local header to after system/libc headers
Suppress unused parameter warning
Change C++ style comment to ISO C90 style comment
Delete extra LF at the end of md4c-html.c


mity:
Thx.

PR #315, opened 2026-05-09
Add footnote reference support

hryhoriiK97:
This PR adds opt-in footnote support via `MD_FLAG_FOOTNOTES`. It recognizes `[^label]` references and `[^label]: ` content definitions, emits only referenced definitions, and orders them by first reference.

It also adds HTML rendering for footnote references and the final footnotes section, including unique backlinks for repeated references. The `md2html` CLI gets a new `--ffootnotes` flag, and the behavior is covered by a new footnote spec file.

Testing covered all test/spec*.txt suites, test/regressions.txt, test/pathological-tests.py, and a 600-second ASan OSS-Fuzz run for fuzz-mdhtml.


src/md4c.h
```diff
@@ -313,6 +327,20 @@ typedef struct MD_SPAN_WIKILINK {
     MD_ATTRIBUTE target;
 } MD_SPAN_WIKILINK_DETAIL;
 
+/* Detailed info for MD_SPAN_FOOTNOTE_REF. */
+typedef struct MD_SPAN_FOOTNOTE_REF_DETAIL {
+    unsigned int index;     /* 1-based sequential number (order of first reference in document) */
```
mity: Is there a reason, why the indexing nature of this is exposed in the interface and its not just an implementation detail?

I mean, if (from application's POV) it serves as a unique identification, wouldn't something like `id` be more self-explanatory?

src/md4c.c
```diff
@@ -1919,6 +1941,184 @@ md_lookup_ref_def(MD_CTX* ctx, const CHAR* label, SZ label_size)
...
+    for(def = ctx->footnote_hashtable[hash % ctx->footnote_hashtable_size];
+        def != NULL;
+        def = def->next)
+    {
+        if(def->hash == hash  &&
+           md_link_label_cmp(def->label, def->label_size, label, label_size) == 0)
+            return def;
+    }
+    return NULL;
+}
```
mity, line 2132: There is a reason why ref. definition hashtable is as complex as it is. It allows to perform binary search in buckets with multiple reference definitions.

If you walk the bucket via a link list, it can be used as a DOS-attack vector with malicious input leading to quadratic parsing times. 

It's quite simple to write a script which generates markdown input with million footnotes all falling into one bucket, and then reference the one at the tail million times.

src/md4c.c
```diff
@@ -1919,6 +1941,184 @@ md_lookup_ref_def(MD_CTX* ctx, const CHAR* label, SZ label_size)
...
+
```
mity, line 2133: It would also be likely better to refactor the code so that the ref. def. hashtable and this footnote hashtable share the code.

E.g. we could have an abstract type hashtable (which assumes some string keys), and then both the reference definitions and the footnotes would reuse that code.

mity (changes requested):
In general, probably the right approach.

However there are some issues which need to be addressed, especially the possible O(n^2) needs to be resolved.

src/md4c.h
hryhoriiK97: Good point. I renamed the public fields from index/ref_index to id/ref_id so the API exposes identifiers rather than implementation indexing - [35f9828](https://github.com/mity/md4c/pull/315/commits/35f9828dd45e15d263189adad2e9744e8c50f9d3)

src/md4c.c
hryhoriiK97: Thanks, updated footnote lookup to use sorted collision buckets with bsearch(), matching ref definitions - [a0cbcd6](https://github.com/mity/md4c/pull/315/commits/a0cbcd67e420c8fd2a6af6811811b242fae77060)

src/md4c.c
hryhoriiK97: Agreed, thanks. I refactored both reference definitions and footnotes to share the same internal hashtable code - [83d3298](https://github.com/mity/md4c/pull/315/commits/83d3298c0c402ef16924c67c8bfd17c2966df593)

hryhoriiK97:
> In general, probably the right approach.
> 
> However there are some issues which need to be addressed, especially the possible O(n^2) needs to be resolved.

@mity thank you for the detailed review. I’ve fixed the reported issues.

mity:
Sorry for the delay, it's caused by some regressions caused by #316 the oss-fuzz keeps reporting. 

I will come back to this when the fuzzing settles down, it's easier for me to deal with such bigger changes one by one.

src/md4c-html.c
```diff
@@ -388,29 +389,77 @@ render_open_wikilink_span(MD_HTML* r, const MD_SPAN_WIKILINK_DETAIL* det)
  ***  HTML renderer implementation  ***
  **************************************/
 
+static void
+render_open_footnote_ref_span(MD_HTML* r, const MD_SPAN_FOOTNOTE_REF_DETAIL* det)
+{
+    char buf[64];
+    if(det->ref_id <= 1)
```
mity: I find somewhat odd we generate
```
<a href="#fnref-1" class="footnote-backref">&#8617;</a>
<a href="#fnref-1-2" class="footnote-backref">&#8617;</a>
```
when there are 2+ references.

Is there any real benefit to treat the 1st reference specially? If not I'd vote for removing this `if` and using the more complex `id` generation unconditionally.

src/md4c-html.c
```diff
@@ -424,22 +473,26 @@ leave_block_callback(MD_BLOCKTYPE type, void* detail, void* userdata)
...
-        case MD_BLOCK_P:            RENDER_VERBATIM(r, "</p>\n"); break;
-        case MD_BLOCK_TABLE:        RENDER_VERBATIM(r, "</table>\n"); break;
-        case MD_BLOCK_THEAD:        RENDER_VERBATIM(r, "</thead>\n"); break;
-        case MD_BLOCK_TBODY:        RENDER_VERBATIM(r, "</tbody>\n"); break;
-        case MD_BLOCK_TR:           RENDER_VERBATIM(r, "</tr>\n"); break;
-        case MD_BLOCK_TH:           RENDER_VERBATIM(r, "</th>\n"); break;
-        case MD_BLOCK_TD:           RENDER_VERBATIM(r, "</td>\n"); break;
+        case MD_BLOCK_DOC:
+            if(r->footnote_count > 0)
+                RENDER_VERBATIM(r, "</ol>\n</section>\n");
```
mity: This is kinda hack which would every custom or 3rd party renderer repeat, so I see this need as a proof of deficiency of the public interface in `md4c.h`.

What about this instead:

1. Add new `MD_BLOCK_FOOTNOTE_DEF_SECTION` (or whatever if you come with better name) into `MD_BLOCK_TYPE`.

2. The parser would then call `enter_block(MD_BLOCK_FOOTNOTE_DEF_SECTION)` and `close_block(MD_BLOCK_FOOTNOTE_DEF_SECTION)` before 1st and after last `MD_BLOCK_FOOTNOTE_DEF` respectively.

3. Renderer would generate `<section><ol>` and `</ol></section>` in response to these new events in very straightforward way.

mity (changes requested):

src/md4c-html.c
hryhoriiK97: Agreed, thanks. I removed the special case and now all footnote references use the same fnref-N-M id format - [b74f3b](https://github.com/mity/md4c/pull/315/commits/2b74f3b468a4de5638d819b2f19b01a51ab96327)

src/md4c-html.c
hryhoriiK97: Thank you for the comment! I added MD_BLOCK_FOOTNOTE_DEF_SECTION, and now the parser wraps emitted footnote definitions with this block, so renderers can open/close the footnote section directly - [7310688](https://github.com/mity/md4c/pull/315/commits/731068884cabd891df137b95f07967c3490f0a15)

hryhoriiK97:
@mity I've addressed your latest comments 🙂

mity (approved):
Thanks, looks good to me.

Issue #306, opened 2026-04-21
#251 not completely fixed, queries with underscore still not recognized


ralfjunker:
Commit 5d52d321f1abd3834c93d451699e2b64cb856180 does not fix the problem with underscores added as a [comment](https://github.com/mity/md4c/issues/251#issuecomment-2266099463) to #251.

The following URLs are valid but still not recognized as of version 0.5.3:

* Query with leading underscore: http://a.b/?_c
* Query with trailing underscore: http://a.b/?c_
* Query with underscore before equals sign: http://a.b/?c_=d
* Query with underscore after equals sign: http://a.b/?c=_d

There might be other combinations.

mity:
Yeah, I've overlooked the comment somehow, sorry for that. I'll take a look at it.

That said, expect very slow and incremental relaxation of the rules in general, I'm quite afraid to make the rules too relaxed in order not to open gates for false positives.

ralfjunker:
Thanks, @mity, for looking at the underscore issue.

I understand you being careful about relaxing URL detection. However, worries for underscores are unwarranted IMO. The HTML spec lists underscore as part of the [URL code points](https://url.spec.whatwg.org/#url-code-points), which make up the [URL-query string](https://url.spec.whatwg.org/#url-query-string).

In conclusion, underscores are allowed anywhere in the URL-query string.

znewt99:
Should the focus for URL detection be RFC 3986 or WHATWG HTML5?

mity:
> Should the focus for URL detection be RFC 3986 or WHATWG HTML5?

Imho, no. Commonmark specification is stricter than that even in recognition of standard [autolink URLs](https://spec.commonmark.org/0.31.2/#absolute-uri) enclosed in `<...>`.

IMHO the permissive autolinks ought to be even stricter, at the very least in the list of supported schemes, and recognize only what clearly is intended as an URL and avoid false positives as much as possible.

Issue #251, opened 2024-03-06
Some character sequences break URL detection


atraczyk:
`/-/` and `/+/` prevents URL detection when calling `md_html`.

Example broken link: `https://codereview.qt-project.org/c/qt/qtwayland/+/545836`

Works fine if enveloped in chevrons: `<https://codereview.qt-project.org/c/qt/qtwayland/+/545836>`

Tested with commit `481fbfb`

Kaffeine:
The same goes for underscore. E.g. `https://www.youtube.com/watch?v=_tEL8bJ7yqU`

Page-MS:
I encountered the exact same issue. 
Is it planned to be fixed ?

PR #319, opened 2026-05-12
Refactor the permissive autolink extension code.

mity:
* Change the main scanning logic so that we don't repeat "is this char a word character" condition in multiple code branches, as it has incrementally diverted from simple alnum() check to a very complex question involving checking different characters for each URL segment, and even whether some of them are not actually part of a resolved mark.

 * Isolate the code shared by the URL segments (username, hostname, path, query and fragment) into new md_analyze_permissive_autolink_segment().

 * That function is able to scan backwards, so it can be reused when scanning for e-mail username (before the '@') instead of (poorly) duplicated code.

 * Having the shared code in a new function, we also get rid of the baroque URL_MAP[] structure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,20 @@ New Features:
 
     Contributed by [Gregory Moskaliuk](https://github.com/hryhoriiK97).
 
+Changes:
+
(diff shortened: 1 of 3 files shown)


mity:
Fixes ~~#251~~ #306, btw (or at the very least those cases mentioned there).

PR #324, opened 2026-05-13
Fix double-free in md_build_attr_append_substr on second realloc failure

Cyberpsych0s1s:
If realloc(substr_types) succeeds and moves the allocation, then realloc(substr_offsets) fails, the old substr_types pointer has already been freed by realloc internally. The error path did not update build->substr_types before returning, so md_free_attribute would later call free() on the dangling old pointer -- a double-free.

to fix we do: build->substr_types = new_substr_types immediately after the first realloc succeeds. The second failure path then needs no explicit free() since md_free_attribute already handles cleanup correctly.


mity:
Good catch.

PR #325, opened 2026-05-14
Fix: Four parser/renderer correctness and OOM bugs.

Cyberpsych0s1s:
Four small independent correctness fixes — full rationale in the commit message.

- `md_build_attr_append_substr` / `md_is_link_reference_definition` :  commit alloc counters only after realloc succeeds (companion to #324).
- `render_utf8_codepoint` :  reject surrogate codepoints (U+D800..U+DFFF).
- `md_is_link_title` :  `\` + newline is not an escape per CommonMark.

Independent of #324 . Full test suite (916 spec + 29 pathological) passes. Happy to split if needed.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -1440,12 +1440,12 @@ md_build_attr_append_substr(MD_CTX* ctx, MD_ATTRIBUTE_BUILD* build,
     if(build->substr_count >= build->substr_alloc) {
         MD_TEXTTYPE* new_substr_types;
         OFF* new_substr_offsets;
-
-        build->substr_alloc = (build->substr_alloc > 0
+        int new_alloc = (build->substr_alloc > 0
                 ? build->substr_alloc + build->substr_alloc / 2
                 : 8);
+
         new_substr_types = (MD_TEXTTYPE*) realloc(build->substr_types,
-                                    build->substr_alloc * sizeof(MD_TEXTTYPE));
+                                    new_alloc * sizeof(MD_TEXTTYPE));
         if(new_substr_types == NULL) {
             MD_LOG("realloc() failed.");
             return -1;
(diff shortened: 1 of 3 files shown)


Cyberpsych0s1s:
my bad I forgot to add test coverage for the changes in the initial commit

src/md4c.c
```diff
@@ -2237,17 +2238,18 @@ md_is_link_reference_definition(MD_CTX* ctx, const MD_LINE* lines, MD_SIZE n_lin
     /* So, it _is_ a reference definition. Remember it. */
     if(ctx->n_ref_defs >= ctx->alloc_ref_defs) {
         MD_REF_DEF* new_defs;
-
-        ctx->alloc_ref_defs = (ctx->alloc_ref_defs > 0
+        int new_alloc = (ctx->alloc_ref_defs > 0
```
mity: Same as for `md_build_attr_append_substr()`.

It's immaterial, but we can accept the change provided we use the right type: `SZ new_alloc`.

mity:
> * `md_build_attr_append_substr` / `md_is_link_reference_definition` :  commit alloc counters only after realloc succeeds (companion to [Fix double-free in md_build_attr_append_substr on second realloc failure #324](https://github.com/mity/md4c/pull/324)).

This is immaterial change: If `realloc()` fails, we abort all parsing and don't access the buffers anymore. We could perhaps accept the change for kind of local correctness and better readability, however then the types of the size variables should be `SZ`.

> * `render_utf8_codepoint` :  reject surrogate codepoints (U+D800..U+DFFF).

Yup, good catch.

> * `md_is_link_title` :  `\` + newline is not an escape per CommonMark.

Actually it is an escape, even though the specification covers it in a dedicated section as a "[hard line break](https://spec.commonmark.org/0.31.2/#hard-line-breaks)".

```
[my long link body with soft line break \
inside of it](http://example.com)
```

Our HTML renderer doesn't really care much about the soft breaks, but some applications with a custom renderer may.

mity:
Merged, however next time please create a new pull request for every fix/feature unless they are closely related.

Cyberpsych0s1s:
thanks, will do!

PR #342, opened 2026-05-22
Fix footnote OSS-Fuzz regression

### LINKED ISSUE #340: OSS-Fuzz issue 515524698
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5236712996667392 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=515524698 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

hryhoriiK97:
Fixes #340 

Fixes a footnote regression found by OSS-Fuzz where certain footnote definitions with bracketed inline content could trigger a debug assertion.
Also updates footnote parsing to treat following paragraph lines as part of the footnote body, matching GitHub-style multiline footnotes, and adds regression/spec coverage for both cases.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -2026,6 +2026,7 @@ md_is_footnote_definition(MD_CTX* ctx, const MD_LINE* lines, MD_SIZE n_lines)
     MD_LINE* content_lines = NULL;
     MD_FOOTNOTE_DEF* def;
     MD_SIZE n;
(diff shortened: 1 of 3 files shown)


mity:
Thanks, that was fast.

PR #349, opened 2026-05-24
Refactor bracket resolution into md_resolve_brackets()

hryhoriiK97:
Refactors bracket span resolution per [discussion](https://github.com/mity/md4c/issues/347#issuecomment-4527711228) on #347 : rename md_resolve_links() → md_resolve_brackets(), split wiki-link / link / footnote logic into helpers, and fold footnote resolution into that entry point (no separate call from md_analyze_inlines()).

Footnotes still run as a second pass inside md_resolve_brackets() so wiki destinations are handled before [^label] refs (e.g. [[[^|]]]).

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -3677,7 +3677,7 @@ md_analyze_bracket(MD_CTX* ctx, int mark_index)
      * link in the right order, from inside to outside in case of nested
      * brackets.
      *
-     * The resolving itself is deferred to md_resolve_links().
+     * The resolving itself is deferred to md_resolve_brackets().
      */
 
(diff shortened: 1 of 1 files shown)


mity:
Oh my, thanks. I didn't expect you to jump at it, I more or less just wanted your ack.

mity:
Just FYI: Looking at the previous code from my side, there are some related stuff I want to change, so I will continue to tinker with it too.

hryhoriiK97:
> Oh my, thanks. I didn't expect you to jump at it, I more or less just wanted your ack.

@mity no problem at all! Just wanted to help 🙂

PR #350, opened 2026-05-24
[WIP] Refactor the bracket handling and link resolving.

mity:
To do (incomplete):

* [ ] `md_resolve_bracket_wikilink()`: Consider whether the no-new-line in wiki-link destination is right or not.
* [ ] Try to distill common code from `md_resolve_bracket_wikilink()` and `md_resolve_bracket_link()` to the caller, especially handling of the `last_link_beg` et al.
* [x] Convert `md_resolve_bracket_footnotes()` to do just one, and call it from the `while` loop in `md_resolve_brackets()`.
* [ ] Take a look at links vs image handling: Even if there is preceding `'!'`, we should likely still try other interpretations (wiki link, footnote) if it is not really an image.
* [x] Try to get rid of the special footnote case in `md_disable_marks()`. This smells more like damage control rather than a proper fix.


mity:
Not complete to my satisfaction, but merging to fix the #352.

PR #360, opened 2026-05-30
Improve git ignore

znewt99:


mity:
This is some general wanna-be kind of universal `.gitignore` for all the world, isn't it?

Quite frankly, I don't want to accept such a thing as imho it could potentially blow into our face. E.g. ignoring `Makefile` because we might hypothetically switch to `autoconf` machinery which then would generate `Makefile`? What if instead we switch to plain old manually written `Makefile` in the future?

I have no problem to accept changes addressing actual build artifacts on more or less any platform, but only actual artifacts with our actual build system.

znewt99:
The current `.gitignore` already includes Makefile in the section that includes autotools machinery.

mity:
I might have used a bad example then, given the comment of Makefile in the diff then.

(And I also admit part of it may be I generally do not rely on the `.gitignore` file, as I build in a subdir which is ignored thanks to my local git configuration during my development.)

Nevertheless a PR which essentially replaces all the file with no explanation whatsoever isn't really a way to go for me.

PR #381, opened 2026-07-30
Add ESP-IDF / PlatformIO / Arduino packaging support

NellowTCS:
Adds packaging metadata so md4c can be consumed as a component/library by embedded build systems, used for my app for the PocketMage (and possibly PocketMageOS later on).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,14 @@ project(MD4C
     LANGUAGES C
 )
 
+if(IDF_TARGET)
+    idf_component_register(
+        SRC_DIRS "src"
+        INCLUDE_DIRS "src"
+    )
+    return() # Stops executing the desktop configurations below when building for ESP32
+endif()
+
 option(BUILD_MD2HTML_EXECUTABLE "Whether to compile the md2html executable" ON)
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,63 +6,68 @@ project(MD4C
     LANGUAGES C
 )
 
+configure_file(idf_component.yml.in    ${CMAKE_CURRENT_SOURCE_DIR}/idf_component.yml    @ONLY)
+configure_file(library.json.in         ${CMAKE_CURRENT_SOURCE_DIR}/library.json         @ONLY)
+configure_file(library.properties.in   ${CMAKE_CURRENT_SOURCE_DIR}/library.properties   @ONLY)
+
 if(IDF_TARGET)
     idf_component_register(
         SRC_DIRS "src"
         INCLUDE_DIRS "src"
     )
-    return() # Stops executing the desktop configurations below when building for ESP32
 endif()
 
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -6,6 +6,14 @@ project(MD4C
     LANGUAGES C
 )
 
+if(IDF_TARGET)
+    idf_component_register(
+        SRC_DIRS "src"
+        INCLUDE_DIRS "src"
+    )
+    return() # Stops executing the desktop configurations below when building for ESP32
```
mity: I quite strongly dislike the `return()`; in case of any future changes to the `CMakeLists.txt` it could lead to unexpected impact on that platform.

Rather wrap the option below with some `if()` to disable it or with some logic to turn it by default off; whatever makes better sense.

idf_component.yml
```diff
@@ -0,0 +1,17 @@
+version: "0.5.3"
```
mity, line 1: Isn't it possible to generate this file via `configure_file()` from `CMakeLists.txt`? If yes, please do so, so that the version is not hard-coded.

library.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "md4c",
+  "version": "0.5.3",
```
mity, line 3: Ditto.

library.properties
```diff
@@ -0,0 +1,11 @@
+name=md4c
+version=0.5.3
```
mity, line 2: Ditto.

mity (changes requested):

CMakeLists.txt
NellowTCS: Good call, agreed. Fixed that!

idf_component.yml
NellowTCS: Done!

Note: I kept the generated files committed as all three build systems (PlatformIO, ESP registry, Arduino) read them before CMake runs, so they need to exist on a fresh checkout.

mity: > Note: I kept the generated files committed as all three build systems (PlatformIO, ESP registry, Arduino) read them before CMake runs, so they need to exist on a fresh checkout.

Ouch, in that case it doesn't make much sense to generate them as it won't save any maintenance work so pls revert those and I'll merge this.

NellowTCS: Ah, my bad, wasn’t clear. Not saying you’d edit them by hand; they just have to exist in the repo on a fresh checkout instead of being made on demand. PlatformIO, ESP registry, and Arduino all read them before CMake runs, so ‎`configure_file()` would generate them too late.

NellowTCS:
> Ah, my bad, wasn’t clear. Not saying you’d edit them by hand; they just have to exist in the repo on a fresh checkout instead of being made on demand. PlatformIO, ESP registry, and Arduino all read them before CMake runs, so ‎`configure_file()` would generate them too late.

Just re-sending in case it was missed!

mity (approved):

PR #385, opened 2026-08-07
Add optional blank line preservation (MD_FLAG_PRESERVEBLANKLINES)

### LINKED ISSUE #384: Feature request: optional flag to preserve extra blank lines
### Summary

Add an **opt-in** parser flag that preserves consecutive blank lines from the source, exposing each *extra* blank line as a new lightweight block type in the callback stream. This is intended for WYSIWYG-style consumers (editors, mobile renderers that build their own view tree from the MD4C callbacks) where the number of blank lines the user typed is meaningful and should survive into the output.

It is deliberately non-CommonMark and therefore gated behind a flag; with the flag off, behaviour and output are unchanged.

### Motivation / use case

We consume the MD4C callback stream directly (not the HTML renderer) to build a native view tree on iOS/Android, aiming for a WYSIWYG-ish experience. In CommonMark, any run of blank lines between two blocks collapses to a single block separator, so a user who typed three empty lines between two paragraphs sees the same result as one empty line - the vertical spacing they intentionally added is lost.

The information is discarded very early: in `md_process_line()`, a blank line simply ends the current block, and consecutive blanks are no-ops. Nothing about the *count* of blank lines ever reaches a callback, so this cannot be recovered by a downstream consumer or a custom renderer - the only workaround today is to pre-process the Markdown string (forcing empty paragraphs) before handing it to the parser, which every consumer has to reimplement and which is easy to get wrong around fenced code, lists, etc.

Exposing it natively (as an opt-in) lets consumers who want WYSIWYG semantics get them without string hacks, while everyone else is unaffected.

### Proposed design

- New flag `MD_FLAG_PRESERVEBLANKLINES` (`0x400000`).
- New block type `MD_BLOCK_BLANK`, appended at the end of `MD_BLOCKTYPE` (so no existing enum values shift). It carries **no contents** (no `MD_TEXT` is emitted) and has `NULL` detail — an `ENTER`/`LEAVE` pair a renderer can turn into a spacer.
- Semantics for a run of *N* consecutive blank lines: the **first** blank keeps its usual role as the CommonMark block separator (not emitted); each **subsequent** blank emits one `MD_BLOCK_BLANK`. So *N* blanks → *N − 1* preserved blocks, matching the empty-paragraph workaround.
- **Leading** blank lines (before any block has been buffered) are suppressed. **Trailing** blanks are preserved. Blanks inside containers (list items, block quotes) nest correctly, because the blank block is pushed into the same buffered block stream as the container open/close markers.

### Example

Input:

```
Paragraph one.



Paragraph two.
```

Callback stream **with** the flag (schematically):

```
ENTER P → TEXT "Paragraph one." → LEAVE P
ENTER BLANK → LEAVE BLANK      (extra blank #1)
ENTER BLANK → LEAVE BLANK      (extra blank #2)
ENTER P → TEXT "Paragraph two." → LEAVE P
```

With the flag **off**: identical to today (one `P`, one `P`).

The bundled `md4c-html` renderer, for demonstration, renders each `MD_BLOCK_BLANK` as `<p> </p>`.

### Verification

- Full CommonMark suite (`test/run-testsuite.py -s test/spec.txt`): **652 passed, 0 failed** with the flag off — default output is byte-for-byte unchanged.
- Verified the new behaviour both at the raw `MD_PARSER` callback level and via `md_html(...)`.
- Sanity-checked edge cases: leading blanks (suppressed), trailing blanks (preserved), blanks between list items, blanks inside block quotes.

### Open questions for maintainers

1. Is an opt-in, non-CommonMark flag like this acceptable in the core, or would you prefer it live only as documentation of the pre-processing approach?
2. Preferred representation: a new `MD_BLOCK_BLANK` type (as here), vs. reusing an empty `MD_BLOCK_P`, vs. a single block carrying a *count* in a detail struct?
3. Leading/trailing-blank policy (currently: suppress leading, keep trailing). Is that the behaviour you'd want?
4. Interaction with list looseness (tight vs. loose) deserves dedicated spec tests before this would be merge-ready; happy to add a `test/spec-preserve-blank-lines.txt` if the direction is welcome.

I'm attaching a working proof-of-concept patch below (against `0.5.3` / 9de89d8). It's ~33 lines across `md4c.h`, `md4c.c`, and `md4c-html.c`. Happy to turn it into a proper PR with tests and CLI wiring if you're open to the feature.

_Note: I'm not sure what's your approach on this, but this patch was AI generated and then verified by me_

<details>
<summary>Proof-of-concept patch (<code>git apply</code>-able)</summary>

```diff
diff --git a/src/md4c-html.c b/src/md4c-html.c
index 590078d..0d11c4d 100644
--- a/src/md4c-html.c
+++ b/src/md4c-html.c
@@ -452,6 +452,7 @@ enter_block_callback(MD_BLOCKTYPE type, void* detail, void* userdata)
         case MD_BLOCK_FOOTNOTE_DEF_SECTION: RENDER_VERBATIM(r, "<section class=\"footnotes\">\n<ol>\n"); break;
         case MD_BLOCK_FOOTNOTE_DEF: render_open_footnote_def_block(r, (MD_BLOCK_FOOTNOTE_DEF_DETAIL*)detail); break;
         case MD_BLOCK_ADMONITION:   render_open_admonition_block(r, (const MD_BLOCK_ADMONITION_DETAIL*) detail); break;
+        case MD_BLOCK_BLANK:    RENDER_VERBATIM(r, "<p>\xc2\xa0</p>\n"); break;
     }

     return 0;
@@ -483,6 +484,7 @@ leave_block_callback(MD_BLOCKTYPE type, void* detail, void* userdata)
         case MD_BLOCK_FOOTNOTE_DEF_SECTION: RENDER_VERBATIM(r, "</ol>\n</section>\n"); break;
         case MD_BLOCK_FOOTNOTE_DEF: render_close_footnote_def_block(r, (MD_BLOCK_FOOTNOTE_DEF_DETAIL*)detail); break;
         case MD_BLOCK_ADMONITION:   RENDER_VERBATIM(r, "</div>\n"); break;
+        case MD_BLOCK_BLANK:    /* noop (whole block emitted on enter) */ break;
     }

     return 0;
diff --git a/src/md4c.c b/src/md4c.c
index 833c074..3a72b5c 100644
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -5502,7 +5502,8 @@ md_process_leaf_block(MD_CTX* ctx, MD_BLOCK* block)
     /* Process the block contents accordingly to is type. */
     switch(block->type) {
         case MD_BLOCK_HR:
-            /* noop */
+        case MD_BLOCK_BLANK:
+            /* noop (no contents) */
             break;

         case MD_BLOCK_CODE:
@@ -6988,6 +6989,26 @@ md_process_line(MD_CTX* ctx, const MD_LINE_ANALYSIS** p_pivot_line, MD_LINE_ANAL
     /* Blank line ends current leaf block. */
     if(line->type == MD_LINE_BLANK) {
         MD_CHECK(md_end_current_block(ctx));
+
+        /* Optionally preserve extra blank lines for WYSIWYG-like consumers.
+         * Within a run of consecutive blank lines, the first blank acts as the
+         * usual CommonMark block separator (not preserved). Every subsequent
+         * blank in the run (pivot is already the dummy blank line) emits one
+         * empty MD_BLOCK_BLANK. Leading blank lines (before any content has been
+         * buffered) are suppressed. */
+        if((ctx->parser.flags & MD_FLAG_PRESERVEBLANKLINES)  &&
+           *p_pivot_line == &md_dummy_blank_line  &&
+           ctx->n_block_bytes > 0)
+        {
+            MD_BLOCK* block = (MD_BLOCK*) md_push_block_bytes(ctx, sizeof(MD_BLOCK));
+            if(block == NULL)
+                return -1;
+            block->type = MD_BLOCK_BLANK;
+            block->flags = 0;
+            block->data = 0;
+            block->n_lines = 0;
+        }
+
         *p_pivot_line = &md_dummy_blank_line;
         return 0;
     }
diff --git a/src/md4c.h b/src/md4c.h
index e649059..8cddd92 100644
--- a/src/md4c.h
+++ b/src/md4c.h
@@ -118,7 +118,14 @@ typedef enum MD_BLOCKTYPE {
     /* Adminition extension.
      * Detail MD_BLOCK_ADMONITION_DETAIL.
      * Note: Recognized only when MD_FLAG_ADMONITIONS is enabled. */
-    MD_BLOCK_ADMONITION
+    MD_BLOCK_ADMONITION,
+
+    /* An extra blank (empty) line preserved from the source, beyond the single
+     * blank line CommonMark uses as a block separator. One such block is emitted
+     * per extra blank line in a run. It carries no contents (no MD_TEXT is sent)
+     * and has NULL detail.
+     * Note: Emitted only when MD_FLAG_PRESERVEBLANKLINES is enabled. */
+    MD_BLOCK_BLANK
 } MD_BLOCKTYPE;

 /* Span represents an in-line piece of a document which should be rendered with
@@ -389,6 +396,7 @@ typedef struct MD_BLOCK_FOOTNOTE_DEF_DETAIL {
 #define MD_FLAG_ADMONITIONS                 0x80000 /* Enable admonitions extension. */
 #define MD_FLAG_FOOTNOTES                   0x100000 /* Enable [^label] footnote references. */
 #define MD_FLAG_HIGHLIGHT                   0x200000 /* Enable ==highlight== spans. */
+#define MD_FLAG_PRESERVEBLANKLINES          0x400000 /* Preserve extra blank lines as MD_BLOCK_BLANK (non-CommonMark; for WYSIWYG). */

 #define MD_FLAG_PERMISSIVEAUTOLINKS         (MD_FLAG_PERMISSIVEEMAILAUTOLINKS | MD_FLAG_PERMISSIVEURLAUTOLINKS | MD_FLAG_PERMISSIVEWWWAUTOLINKS)
 #define MD_FLAG_NOHTML                      (MD_FLAG_NOHTMLBLOCKS | MD_FLAG_NOHTMLSPANS)
```

</details>

eszlamczyk:
Closes #384.

## Motivation

When rendering from the MD4C callback stream (e.g. building a native view tree for a WYSIWYG-like editor), the number of blank lines the author typed between blocks is meaningful. CommonMark collapses any run of blank lines into a single block boundary, and MD4C discards this information in the process so it can't be recovered downstream without pre-processing the source. This adds an opt-in way to get that information from the parser.

## What this does

Adds the flag `MD_FLAG_PRESERVEBLANKLINES`. When enabled, each run of blank lines separating two blocks is reported as a single new block `MD_BLOCK_BLANK`, whose detail `MD_BLOCK_BLANK_DETAIL` carries the number of blank lines:

```c
  typedef struct MD_BLOCK_BLANK_DETAIL {
      unsigned count;   /* Count of blank lines forming the block separation */
  } MD_BLOCK_BLANK_DETAIL;
```

Following the discussion in #384, the run is coalesced into one block (the count is stored once, not one block per line), and all blank lines are reported - i.e. the arser reports "this block separation is made of N blank lines" and leaves the rendering decision to the application.

With the flag off, behaviour and output are byte-for-byte unchanged.

## Semantics

- One `MD_BLOCK_BLANK` per run; no `MD_TEXT` is sent.
- Blank lines attach to the innermost open container and flush at the next block boundary, so runs split correctly across edges: blanks inside a block quote are reported inside it, blanks after it at the outer level (and vice versa).
- Only genuinely empty lines are counted. Blank lines inside code blocks are verbatim content, and a few non-blank lines MD4C internally treats as blank (a closing code fence, an admonition tag) are excluded.

md2html gains --fpreserve-blank-lines; its renderer emits count non-breaking-space paragraphs so the count is visible/testable.

*Note: Blank lines trailing a list's last item (before a dedented paragraph that ends the list) currently attach inside that last `<li>`, consistent with the "innermost open container" rule. Happy to tweak those semantics, if you'd prefer. Documented with a test in `test/coverage.txt`*

### Tests & docs

- New test/spec-preserve-blank-lines.txt (10 examples, doubling as docs) plus harder paths in test/coverage.txt (closing-fence exclusion, HTML-end blank, loose lists, trailing-in-item, nested containers, whitespace-only lines).
- Full suite green; -Werror Debug + Release clean. `CRLF` verified manually but not added as a spec test (the .txt files are LF-only).
- README, CHANGELOG, md2html man page, and md4c.h doc comments updated.

### DIFF
--- a/md2html/md2html.c
+++ b/md2html/md2html.c
@@ -325,7 +325,7 @@ usage(void)
         "                       Same as --fpermissive-email-autolinks --fpermissive-url-autolinks\n"
         "                       --fpermissive-www-autolinks\n"
         "      --fpreserve-blank-lines\n"
-        "                       Report blank line runs (as MD_BLOCK_BLANK)\n"
+        "                       Render extra blank lines between blocks as <br>\n"
         "      --fhighlight    Enable highlight spans (==text==)\n"
         "      --fspoilers      Enable spoiler spans (||hidden text||)\n"
         "      --fstrikethrough Enable strike-through spans\n"
--- a/md2html/md2html.c
+++ b/md2html/md2html.c
@@ -262,6 +262,7 @@ static const CMDLINE_OPTION cmdline_options[] = {
     {  0,  "fpermissive-email-autolinks",   '@', 0 },
     {  0,  "fpermissive-url-autolinks",     'U', 0 },
     {  0,  "fpermissive-www-autolinks",     '.', 0 },
+    {  0,  "fpreserve-blank-lines",         'Y', 0 },
     {  0,  "fspoilers",                     'P', 0 },
     {  0,  "fstrikethrough",                'S', 0 },
     {  0,  "fsubscripts",                   '~', 0 },
@@ -323,6 +324,8 @@ usage(void)
         "      --fpermissive-autolinks\n"
         "                       Same as --fpermissive-email-autolinks --fpermissive-url-autolinks\n"
         "                       --fpermissive-www-autolinks\n"
+        "      --fpreserve-blank-lines\n"
+        "                       Report blank line runs (as MD_BLOCK_BLANK)\n"
         "      --fhighlight    Enable highlight spans (==text==)\n"
         "      --fspoilers      Enable spoiler spans (||hidden text||)\n"
         "      --fstrikethrough Enable strike-through spans\n"
@@ -413,6 +416,7 @@ cmdline_callback(int opt, char const* value, void* data)
         case '_':   parser_flags |= MD_FLAG_UNDERLINE; break;
         case 'N':   parser_flags |= MD_FLAG_FOOTNOTES; break;
         case 'B':   parser_flags |= MD_FLAG_HARD_SOFT_BREAKS; break;
+        case 'Y':   parser_flags |= MD_FLAG_PRESERVEBLANKLINES; break;
 
         default:
             fprintf(stderr, "Illegal option: %s\n", value);
--- a/src/md4c-html.c
+++ b/src/md4c-html.c
@@ -426,6 +426,14 @@ render_close_footnote_def_block(MD_HTML* r, const MD_BLOCK_FOOTNOTE_DEF_DETAIL* 
     RENDER_VERBATIM(r, "\n</li>\n");
 }
 
+static void
+render_blank_block(MD_HTML* r, const MD_BLOCK_BLANK_DETAIL* det)
+{
+    unsigned i;
+    for(i = 0; i < det->count; i++)
+        RENDER_VERBATIM(r, "<p>&nbsp;</p>\n");
+}
+
 static int
 enter_block_callback(MD_BLOCKTYPE type, void* detail, void* userdata)
 {
@@ -452,6 +460,7 @@ enter_block_callback(MD_BLOCKTYPE type, void* detail, void* userdata)
         case MD_BLOCK_FOOTNOTE_DEF_SECTION: RENDER_VERBATIM(r, "<section class=\"footnotes\">\n<ol>\n"); break;
         case MD_BLOCK_FOOTNOTE_DEF: render_open_footnote_def_block(r, (MD_BLOCK_FOOTNOTE_DEF_DETAIL*)detail); break;
         case MD_BLOCK_ADMONITION:   render_open_admonition_block(r, (const MD_BLOCK_ADMONITION_DETAIL*) detail); break;
+        case MD_BLOCK_BLANK:    render_blank_block(r, (const MD_BLOCK_BLANK_DETAIL*) detail); break;
     }
 
     return 0;
@@ -483,6 +492,7 @@ leave_block_callback(MD_BLOCKTYPE type, void* detail, void* userdata)
         case MD_BLOCK_FOOTNOTE_DEF_SECTION: RENDER_VERBATIM(r, "</ol>\n</section>\n"); break;
         case MD_BLOCK_FOOTNOTE_DEF: render_close_footnote_def_block(r, (MD_BLOCK_FOOTNOTE_DEF_DETAIL*)detail); break;
         case MD_BLOCK_ADMONITION:   RENDER_VERBATIM(r, "</div>\n"); break;
+        case MD_BLOCK_BLANK:    /* noop (fully emitted on enter) */ break;
     }
 
     return 0;
--- a/src/md4c-html.c
+++ b/src/md4c-html.c
@@ -430,8 +430,10 @@ static void
 render_blank_block(MD_HTML* r, const MD_BLOCK_BLANK_DETAIL* det)
 {
     unsigned i;
-    for(i = 0; i < det->count; i++)
-        RENDER_VERBATIM(r, "<p>&nbsp;</p>\n");
+    /* The first blank line is the ordinary block separation CommonMark already
+     * implies; render only the surplus (line_count - 1) as <br>. */
+    for(i = 1; i < det->line_count; i++)
+        RENDER_VERBATIM(r, (r->flags & MD_HTML_FLAG_XHTML) ? "<br/>\n" : "<br>\n");
 }
 
 static int
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -5497,7 +5497,7 @@ md_process_leaf_block(MD_CTX* ctx, MD_BLOCK* block)
             break;
 
         case MD_BLOCK_BLANK:
-            det.blank.count = block->data;
+            det.blank.line_count = block->data;
             break;
 
         default:
@@ -5865,9 +5865,7 @@ md_add_line_into_current_block(MD_CTX* ctx, const MD_LINE_ANALYSIS* analysis)
     return 0;
 }
 
-/* Report any pending blank lines as a single MD_BLOCK_BLANK. Called just before
- * a leaf or container block is pushed, so the blank block lands on the correct
- * side of the boundary. No-op unless MD_FLAG_PRESERVEBLANKLINES is enabled. */
+/* Part of the MD_FLAG_PRESERVEBLANKLINES implementation. */
 static int
 md_flush_blank_lines(MD_CTX* ctx)
 {
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -275,6 +275,10 @@ struct MD_CTX_tag {
     int n_block_bytes;
     int alloc_block_bytes;
 
+    /* Pending count of blank lines not yet reported as MD_BLOCK_BLANK.
+     * Used only with MD_FLAG_PRESERVEBLANKLINES. */
+    unsigned n_blank_lines;
+
     /* For container block analysis. */
     MD_CONTAINER* containers;
     int n_containers;
@@ -5432,6 +5436,7 @@ md_process_leaf_block(MD_CTX* ctx, MD_BLOCK* block)
         MD_BLOCK_H_DETAIL header;
         MD_BLOCK_CODE_DETAIL code;
         MD_BLOCK_TABLE_DETAIL table;
+        MD_BLOCK_BLANK_DETAIL blank;
     } det;
     MD_ATTRIBUTE_BUILD info_build = { 0 };
     MD_ATTRIBUTE_BUILD lang_build = { 0 };
@@ -5491,6 +5496,10 @@ md_process_leaf_block(MD_CTX* ctx, MD_BLOCK* block)
             det.table.body_row_count = block->n_lines - 2;
             break;
 
+        case MD_BLOCK_BLANK:
+            det.blank.count = block->data;
+            break;
+
         default:
             /* Noop. */
             break;
@@ -5502,7 +5511,8 @@ md_process_leaf_block(MD_CTX* ctx, MD_BLOCK* block)
     /* Process the block contents accordingly to is type. */
     switch(block->type) {
         case MD_BLOCK_HR:
-            /* noop */
+        case MD_BLOCK_BLANK:
+            /* noop (no contents) */
             break;
 
         case MD_BLOCK_CODE:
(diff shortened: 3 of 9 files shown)


md2html/md2html.c
```diff
@@ -323,6 +324,8 @@ usage(void)
         "      --fpermissive-autolinks\n"
         "                       Same as --fpermissive-email-autolinks --fpermissive-url-autolinks\n"
         "                       --fpermissive-www-autolinks\n"
+        "      --fpreserve-blank-lines\n"
+        "                       Report blank line runs (as MD_BLOCK_BLANK)\n"
```
mity: This describes what `md2html` utility does also to **end users**, who are just interested to convert an input to HTML and who possibly know nothing about implementation details or parser internals. 

So it needs to be reworded, e.g. to `Output blank lines as <br>` or something such user can understand.

src/md4c-html.c
```diff
@@ -426,6 +426,14 @@ render_close_footnote_def_block(MD_HTML* r, const MD_BLOCK_FOOTNOTE_DEF_DETAIL*
     RENDER_VERBATIM(r, "\n</li>\n");
 }
 
+static void
+render_blank_block(MD_HTML* r, const MD_BLOCK_BLANK_DETAIL* det)
+{
+    unsigned i;
+    for(i = 0; i < det->count; i++)
+        RENDER_VERBATIM(r, "<p>&nbsp;</p>\n");
```
mity: Wouldn't `<br>` (or `<br/>` in XHTML mode) be more appropriate?

Also I wonder whether the same logic (N-1) shouldn't be used here as in your renderer. (Yes it takes it further from parser's logic but again, this utility serves dual purpose: our testing needs as well as end users.)

src/md4c.h
```diff
@@ -363,6 +368,11 @@ typedef struct MD_BLOCK_FOOTNOTE_DEF_DETAIL {
     MD_ATTRIBUTE label;     /* Raw label text */
 } MD_BLOCK_FOOTNOTE_DEF_DETAIL;
 
+/* Detailed info for MD_BLOCK_BLANK. */
+typedef struct MD_BLOCK_BLANK_DETAIL {
+    unsigned count;         /* Count of blank lines forming the block separation */
```
mity: Please rename the member to something more descriptive, e.g. `line_count`.

src/md4c.c
```diff
@@ -5855,13 +5865,38 @@ md_add_line_into_current_block(MD_CTX* ctx, const MD_LINE_ANALYSIS* analysis)
     return 0;
 }
 
+/* Report any pending blank lines as a single MD_BLOCK_BLANK. Called just before
+ * a leaf or container block is pushed, so the blank block lands on the correct
+ * side of the boundary. No-op unless MD_FLAG_PRESERVEBLANKLINES is enabled. */
+static int
```
mity, line 5869: Make the comment shorter, e.g. just mention it is part of `MD_FLAG_PRESERVEBLANKLINES` implementation, or get it out completely.

If the logic changes, the comment would become outdated too easily. Also unless I missed some logic in the PR I think it's not true as `block_push_bytes()` is also called from `md_enter_container_blocks()`; i.e. it's not only leaf blocks, it's all blocks.

And it's likely the right thing to do, in order to handle nesting of blank lines at a beginning of a block quote.

However `md_flush_blank_lines()` should be probably also called from `md_leave_child_containers()`, at the very least for block quote.

Consider this input:
```
> block quote with few trailing blank lines **inside** of it.
> 
>


And then additional blank lines **after** the block quote.
```

We should honor the proper nesting of the blank lines.

mity (changes requested):
Looks as the right approach to me. There are some requested changes though.

src/md4c-html.c
eszlamczyk: Went with the N-1 count as you suggested. 

_Note: this needed changing the tests_

src/md4c.c
eszlamczyk: @mity 
Shortened the comment to just note it's part of the `MD_FLAG_PRESERVEBLANKLINES` implementation.

On the flush: after empirical tests, it turns out the nesting is already honored, so I kept the single flush rather than adding one to `md_leave_child_containers()`. That function closes each container via `md_push_container_bytes()`, which already calls `md_flush_blank_lines()` first - so blanks flush on the correct side of every container closer. Your example gives exactly what you'd want:

```
> block quote with trailing blank lines inside of it.
>
>


And then blank lines after the block quote.
```
```
ENTER QUOTE
  ENTER P … LEAVE P
  BLANK { line_count = 2 }      ← the two ">"-marked blanks, inside the quote
LEAVE QUOTE
BLANK { line_count = 2 }        ← the two unmarked blanks, at document level
ENTER P … LEAVE P
```

`>`-marked blanks report inside, unmarked ones after; 3-level nesting and quote→list splits verify too. Adding a second flush in `md_leave_child_containers()` would double-report that inside run. I've added your example as a test in `test/coverage.txt`.

Separately, while testing this I found empty list items (`-`, `*`, `1.` with no content) were being miscounted as blank separations (e.g. `-\n-\n` reported two phantom blanks) - they're internally retyped to `MD_LINE_BLANK`. Fixed by excluding lines that carry a container mark, plus a test.

One thing I'd like your call on though. Since it's the same "honor proper nesting" theme: block quotes and lists differ for an *unmarked* blank that trails a container. A block quote is left the moment an unmarked line appears (on the blank line itself), so trailing blanks pop out to the outer level. A list stays open across blanks for loose-list detection, so a blank trailing the last item currently attaches *inside* that last `<li>`:

```
- 1
- 2

test
```
```
UL
  LI_START "1" LI_END
  LI_START "2"
    BLANK { line_count = 1 }    ← currently inside the last item
    LI_END
UL end
P "test"
```

By the block-quote logic that unmarked blank arguably belongs at the outer level instead (between `</ul>` and `<p>test</p>`). Do you want lists to match block quotes (pop the trailing blank out), or keep the innermost-container behavior I documented in the PR? I can implement the pop-out - at list-close time the only pending blank is the trailing one (loose-internal blanks already flushed via the next item's opener), so it's tractable without disturbing the internal case.

mity: Ok

mity (approved):

PR #390, opened 2026-08-26
Replace TRUE and FALSE macros with stdbool.h

znewt99:


mity:
Yeah, it might be the right time to do it, hopefully or relevant platforms already have C99 and consequently `<stdbool.h>` nowadays. If not, we may later resolve it by some preprocessor magic if anybody repots problem on some platform.

PR #392, opened 2026-08-27
Build improvements

mity:
* Separate heuristics of the default behavior from actual `option()` declaration.

 * Added new option `BUILD_HTML_LIBRARY` allowing to enable/disable build of our HTML renderer library.

 * By default set `BUILD_MD2HTML_EXECUTABLE` to `ON` only when we're being built as a top-level project. (I assume when another project embeds us, it's usually only interested in our lib target(s)).

 * Abort with an error when the set options cannot be adhered due to their inter-dependency.

 * For the sake of consistency, move logic of adhering the option `BUILD_MD2HTML_EXECUTABLE` from the top-level `CMakeLists.txt` to `md2html/CMakeLists.txt`.

 * Change handling of the `idf_components.yml`, `library.json` and `library.properties` files to a "configure-time assertion" instead of rewriting the pre-generated files in the source directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.14)
 
 project(MD4C
     VERSION "0.5.3"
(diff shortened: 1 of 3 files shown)


mity:
@NellowTCS Can you pls check that this PR (and in particular the commit ~~[be02a9a](https://github.com/mity/md4c/pull/392/commits/be02a9afef881139f32b9080341bf9260d4f4bcf)~~ [c6aab8a](https://github.com/mity/md4c/commit/c6aab8a2044fc3be29e74d46fe50fec09d26e787)) doesn't break build on those embedded platforms?

PR #393, opened 2026-08-29
Fix a -Wpedantic warning related to MD_FALLTHROUGH.

znewt99:
Removes the warning below when -Wpedantic flag is used.

warning: a label can only be part of a statement and a declaration is not a statement [-Wpedantic]

### DIFF
--- a/src/md4c-html.c
+++ b/src/md4c-html.c
@@ -47,9 +47,9 @@
 
 /* For falling through case labels in switch statements. */
 #if defined __clang__ && __clang_major__ >= 12
-    #define MD_FALLTHROUGH()        __attribute__((fallthrough))
+    #define MD_FALLTHROUGH()        ;__attribute__((fallthrough))
 #elif defined __GNUC__ && __GNUC__ >= 7
-    #define MD_FALLTHROUGH()        __attribute__((fallthrough))
+    #define MD_FALLTHROUGH()        ;__attribute__((fallthrough))
 #else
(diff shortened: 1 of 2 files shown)


mity:
Hmmm, I really dislike this, everyone using the macro in the future would tend to forget the braces. Which compiler and what branch of the macro definition causes this?

If it's the fallback 
``` C
#define MD_FALLTHROUGH()        ((void)0)
```
then changing it into a noop-statement would imho be far better fix, e.g.:
``` C
#define MD_FALLTHROUGH()        do {} while(0)
```

znewt99:
It's GCC (`__attribute__((fallthrough))`), when used with the `-Wpedantic` flag that causes the warning when fallthrough is used with cases in switch statements.

Another option is to switch back to `/* FALLTHROUGH */` comments for fallthrough in switch statements. Or, just ignore the warning, and close this PR without merging, since, it's a warning that only appears when the GCC Wpedantic flag is used.

mity:
Ok, finally reproduced. My local gcc 16.1. didn't warn because by default it uses the C23 standard which doesn't require the label to be followed by the statement. It emits the warning if I add e.g. `-Wpedantic --std=c17`.

As ugly as it is, the semicolon before the attribute in the macro definition stops the warning:
```
#define MD_FALLTHROUGH()        ; __attribute__((fallthrough))
```

But at least the ugliness is on the macro definition side, not when using it.

znewt99:
The semicolon looks like the better cleaner option.

PR #404, opened 2026-09-01
Remove #define inline

### LINKED ISSUE #401: Is #define inline still needed?

znewt99:
If you want to see how many complaints you will get. May cause problems for Qt for INTEGRITY. md4c is optionally used in QTextDocument, which is a class within the Qt GUI module. The Qt GUI module is a supported module for Qt for INTEGRITY.

It looks like the current Green Hills Software compiler version does support the inline keyword, but I'm not really sure.

@ec1oud do you know if this will cause a problem?

Closes: #401

### DIFF
--- a/src/md4c-html.c
+++ b/src/md4c-html.c
@@ -30,17 +30,6 @@
 #include "entity.h"
 
 
-#if !defined(__STDC_VERSION__) || __STDC_VERSION__ < 199409L
-    /* C89/90 or old compilers in general may not understand "inline". */
-    #if defined __GNUC__
-        #define inline __inline__
-    #elif defined _MSC_VER
-        #define inline __inline
-    #else
-        #define inline
-    #endif
-#endif
-
 #ifdef _WIN32
     #define snprintf _snprintf
 #endif
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -38,17 +38,6 @@
  ***  Miscellaneous Stuff  ***
  *****************************/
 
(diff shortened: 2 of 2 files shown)


ec1oud:
I don't know much about compilers on Integrity.  But Qt keeps bumping minimum C++ versions; I doubt we have to worry about compilers _that_ old.  But keeping md4c as portable as possible is also a good goal, independent of Qt.

mity:
Well, lets see. There's reason why `git` has the `revert` command, after all. :-)

ec1oud:
FWIW Qt still builds with this change on Linux.

OTOH there have been some changes to the definition of github dialect: admonitions and footnotes.  Have to see if there's anything we can do with that, at some point.

mity:
> FWIW Qt still builds with this change on Linux.

Well, mainstream compilers (including Linux) is surely not where I'd expect any problems with removal of some pre-C99 compatibility hacks.

> OTOH there have been some changes to the definition of github dialect: admonitions and footnotes. Have to see if there's anything we can do with that, at some point.

Sorry you've lost me here, what did github change again in admonitions or footnotes? Did I miss something? Or do you mean what you on Qt side can do about the new features in MD4C?

ec1oud:
> Sorry you've lost me here, what did github change again in admonitions or footnotes? Did I miss something? Or do you mean what you on Qt side can do about the new features in MD4C?

Yes, the latter.  https://qt-project.atlassian.net/browse/QTBUG-149921

PR #407, opened 2026-09-08
Add <video> to block-level HTML tag list (type 6)

hryhoriiK97:
## Summary

The `<video>` element is widely used in markdown to embed video content but is not currently recognized as a block-level HTML tag. Without type 6 recognition, `<video>` falls back to type 7, which does not allow any content after the tag on the same line.

Adding `video` to the type 6 tag list allows `<video src="url"></video>` to be parsed as an HTML block regardless of what follows the opening tag on the same line — consistent with how other media-related tags like `<iframe>` and `<track>` are already handled.

## Changes

- **`src/md4c.c`** — Added `v6[]` array with `"video"` and updated `map6[21]` (`'v'` slot) to point to it.
- **`test/spec-video-tag.txt`** — New test file with 6 examples covering basic recognition, attributes, trailing content, blockquote and list contexts.
- **`CHANGELOG.md`** — Added entry under "Next Version".

`test/normalize.py` already includes `video` in `is_block_tag()` — no change needed.

Reported to the CommonMark spec project: [commonmark/commonmark-spec#841](https://github.com/commonmark/commonmark-spec/issues/841).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## Next Version (Work in Progress)
 
+Changes:
+
+  * Add `<video>` to the list of block-level HTML tags (type 6).
+
 New Features:
 
   * Add blank line preservation, enabled with the flag
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -6187,6 +6187,7 @@ static const TAG s6[] = { X("search"), X("section"), X("summary"), Xend };
 static const TAG t6[] = { X("table"), X("tbody"), X("td"), X("tfoot"), X("th"),
                           X("thead"), X("title"), X("tr"), X("track"), Xend };
 static const TAG u6[] = { X("ul"), Xend };
+static const TAG v6[] = { X("video"), Xend };
 static const TAG xx[] = { Xend };
 
 #undef X
@@ -6202,7 +6203,7 @@ md_is_html_block_start_condition(MD_CTX* ctx, OFF beg)
      * tree to speed-up the search. */
     static const TAG* map6[26] = {
         a6, b6, c6, d6, xx, f6, xx, h6, i6, xx, xx, l6, m6,
-        n6, o6, p6, xx, xx, s6, t6, u6, xx, xx, xx, xx, xx
+        n6, o6, p6, xx, xx, s6, t6, u6, v6, xx, xx, xx, xx
     };
     OFF off = beg + 1;
     int i;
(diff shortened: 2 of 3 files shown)


mity:
Hmmm, sorry, I certainly won't be accepting any changes to the file `spec.txt` whatsoever. It's copied from upstream at https://github.com/commonmark/commonmark-spec and it's very likely the closest thing to any formal widely recognized Markdown specification document with extensive test suite out there, and I won't willy-nilly create or maintain a fork of it.

Any change proposals, bug reports and pull requests to that file should go there.

That said, we might accept small additions in default behavior as the one in this PR even without changing the specification as a kind of an always-enabled extension, assuming it's small/trivial change without far-reaching syntax modifications, adding an explicit option to enable/disable it would be an overkill, any potential incompatibility is obviously outweighed by its usefulness and unlikely to break anything in the real world, and also assuming it doesn't break any test case the specification does explicitly cover.

For me this PR covers all those checkboxes so we might create a new test file for changes of this kind.

(But even then please report it to the spec project so that next version may cover it and reduce the dialect babel in the long run.)

hryhoriiK97:
> Hmmm, sorry, I certainly won't be accepting any changes to the file `spec.txt` whatsoever. It's copied from upstream at https://github.com/commonmark/commonmark-spec and it's very likely the closest thing to any formal widely recognized Markdown specification document with extensive test suite out there, and I won't willy-nilly create or maintain a fork of it.
> 
> Any change proposals, bug reports and pull requests to that file should go there.
> 
> That said, we might accept small additions in default behavior as the one in this PR even without changing the the specification as a kind of an always-enabled extension, assuming it's small/trivial change without far-reaching syntax modifications, adding an explicit option to enable/disable it would be an overkill, any potential incompatibility is obviously outweighed by its usefulness and unlikely to break anything in the real world, and also assuming it doesn't break any test case the specification does explicitly cover.
> 
> For me this PR checks all those checkboxes so we might create a new test file for changes of this kind.
> 
> (But even then please report it to the spec project so that next version may cover it covers it and reduce the dialect babel in the long run.)

@mity thanks for the feedback! I've updated the PR:

- Reverted spec.txt - left it untouched as a CommonMark mirror
- Moved tests to a new spec-video-tag.txt file, following the pattern of other spec-*.txt extensions

I also opened an issue on [commonmark/commonmark-spec](https://github.com/commonmark/commonmark-spec/issues/841) to propose adding <video>.

PR #416, opened 2026-09-11
md_analyze_line: Handle blank lines in a list item in less hacky way.

### LINKED ISSUE #413: List rendered tight when item directly contains a thematic break and a paragraph separated by a blank line
Difference in output between md4c and cmark involving tight vs loose lists

input:
``` markdown
-
  ---

  b
```

cmark output:
``` html
<ul>
<li>
<hr>
<p>b</p>
</li>
</ul>
```

md4c output:
``` html
<ul>
<li>
<hr>
</li>
</ul>
<p>b</p>
```

mity:
Fixes #413.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,9 @@ Fixes:
     Big thanks especially to [Ilia Alshanetsky](https://github.com/iliaal) for
     detecting and fixing all of these.
 
+  * [#413](https://github.com/mity/md4c/pull/413):
+    Fix a case of incorrect ending of a list.
+
 
 ## Version 0.5.3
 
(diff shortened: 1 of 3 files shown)


mity:
@NellowTCS Hmm, can you pls check why the embedded builds are failing now? Is it some temporary issue with some remote resources or what?

NellowTCS:
Try a rerun, seems they were cancelled for some reason

mity:
Ahhh it seems to be the setting in `ci-embedded.yml`:

```
concurrency:
  group: build-embedded
  cancel-in-progress: true
```

Is there a particular reason why it's needed? Sometimes people may work on multiple PRs at the same time, or they may push multiple commits in short succession - and then the 2nd one would get canceled (even though the last run one would be more of interest than the preceding runs).

More info: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

NellowTCS:
I had added that to cancel _previous_ ones, as embedded builds are a bit slow, but yeah, removing it is fine!
Edit: especially b/c it isn't working correctly it seems

mity:
Well, the fuzzing workflow takes 10+ minutes so most likely that one will always be the bottleneck.

PR #419, opened 2026-09-12
Rework pipe handling

mity:
This PR addresses issues raised in the PR #412 in an alternative fashion. Consider it proof of concept for now.

It reworks pipe handling in order to better resolve conflicts between table parsing, wiki-link parsing and spoilers as all those use pipe `|` as part of their syntax.

In particular:

 * Consecutive sequences of `|` characters are now coalesced into one `MD_MARK` instance.

 * Wiki-links now only consider spans of length 1 as the delimiter between its target and display label.

 * Spoilers only considers spans of length 2.

 * Tables are willing to use spans of any length as multiple empty table cells for compatibility with GFM with one exception:

   If both tables and spoilers are enabled, table ignores spans of length 2 and keeps them available for the spoilers. However if this proves problematic we might change it so that spoilers cannot be used inside table, and it would continue to mark an empty cell. If spoilers are not enabled, tables continue to use  the spans of length 2 as before.

To do:
 * [ ] This needs some testing.
 * [ ] Decide whether to keep the table exemption for `||` when spoilers are enabled (the easiest alternative is to not support spoilers inside a table).

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -3528,21 +3528,22 @@ md_collect_marks(MD_CTX* ctx, const MD_LINE* lines, MD_SIZE n_lines, int table_m
                 continue;
             }
 
-            /* A potential spoiler delimiter: || ... ||
-             * Checked before the single-| handler so a double pipe is consumed
-             * as one mark and does not become two cell boundaries. */
-            if(ch == _T('|') && (ctx->parser.flags & MD_FLAG_SPOILERS)) {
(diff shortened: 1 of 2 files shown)


mity:
cc @youdie006 Any comments/suggestion/feedback about this? Can you check whether it does resolve the issues for you?

mity:
Merging so it gets good coverage from the oss fuzzer. The spoilers in tables are documented as experimental so I would have no scruples to change that in any later version.

PR #421, opened 2026-09-12
Footnotes: Fix checking for validity of a footnote reference.

### LINKED ISSUE #380: OSS-Fuzz issue 537183305
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5754432558792704 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=537183305 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

mity:
* Moved validation of the label into a separate function so it can be called from reference resolving code.

 * Limit the maximal label negth to 75 chars (same as Github).

 * Verify validity of the label before calling the potentially expensive md_lookup_footnote_def().

   This fixes #380.

### DIFF
--- a/src/md4c.c
+++ b/src/md4c.c
@@ -2018,6 +2018,26 @@ struct MD_FOOTNOTE_DEF_tag {
     MD_SIZE n_content_lines;
 };
 
+static int
+md_is_footnote_label(MD_CTX* ctx, OFF beg, OFF* p_end)
+{
+    OFF end = beg;
+
+    /* No whitespace, no nested square brackets and length below 75 characters.
+     * See https://github.com/mity/md4c/issues/380 */
+    while(end < ctx->size  &&  end - beg <= 75  &&
+            !ISWHITESPACE(end)  &&  !ISANYOF2(end, _T('['), _T(']')))
+        end++;
+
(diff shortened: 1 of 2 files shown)


mity:
Todo:
 * [ ] Consider disabling the footnotes references when nested inside an outer `[^` ... `]` pair, for compatibility with Github.
 * [x] Add it into `pathological_tests.py`.

mity:
Before (with the test case from #380):
``` bash
$ time ./md2html --ffootnotes poc > /dev/null

real    0m22.608s
user    0m0.000s
sys     0m0.000s
```

After with this applied:
``` bash
$ time ./md2html --ffootnotes poc > /dev/null

real    0m0.060s
user    0m0.000s
sys     0m0.015s
```

mity:
Ok, it's in, without changing the nesting behavior: That has much lower priority and apparently [Github has some bugs in footnotes](https://github.com/micromark/micromark-extension-gfm-footnote#bugs) so it doesn't make much sense to copy what actually might be a bug or care about too obscure situations.

PR #425, opened 2026-09-13
Replace `MD_ASSERT` with `assert`

znewt99:
Replacing MD_ASSERT with assert

Also, removes unused but set variable `prev_mark`, it's is only used within asserts, hasn't been actually used since aeddaf5


mity:
This will have to wait after the upcoming release. Also to keep the behavior, `NDEBUG` should be defined in non-debug builds to make the `assert` noop.

znewt99:
CMAKE automatically defines NDEBUG for RELEASE, MINSIZEREL, & RELWITHDEBINFO builds

mity:
> CMAKE automatically defines NDEBUG for RELEASE, MINSIZEREL, & RELWITHDEBINFO builds

Not so few apps just take md4c.h and md4c.c and incorporate them into their build system. I think the sane default for those is to have it without the assertions.

N-R-K:
The current setup of using `__builtin_unreachable` (and similar) is better for optimization hints in theory. Standard assert macro doesn't have the same behaviour in that regard.

PR #426, opened 2026-09-13
md_consume_link_reference_definitions: Fix manipulation...

mity:
of `ctx->block_bytes[]` if we inject the `MD_BLOCK_HR`.

Addresses #423.


mity:
Ouch, seems Github has some partial outage of pull requests: https://www.githubstatus.com/incidents/0rn90wk115q9

PR #433, opened 2026-09-21
Combine superscript, highlight, and insert portions of md_collect_marks

znewt99:
Looks like there is also a way to combine some of the `md_analyze_` functions, but I haven't figure out a way that makes sense.


mity:
I'm afraid the deduplication of the analyze functions is gonna need a bigger refactorization. My very rough[^1] plan goes along these lines, if you want to help with it:

[^1]: i.e. there may be obstacles and dragons on the path I don't foresee.

1. Write a quite simple generic `md_analyze_generic()` or `md_analyze_simple_mark()` or whatever, which would take two more arguments: the character making the syntax (or possible even two: opener and closer, really simple marks could use same char for both) and a pointer to its respective opener chain as function parameter.

2. Change `md_analyze_XXX()` of some simple marks to be just a simple wrapper of this new function.

3. Review if possible to go back to step 1 and by doing some reasonable amount of generalization make it able to absorb more `md_analyze_XXX()` functions.

4. Create some data mapping/data structure which ties together: 1) the flag enabling the feature (or 0 if in CommonMark), 2) the character(s) making the mark and 3) the opener chain; then we might get rid of those `md_analyze_XXX()` for simple marks completely as it could be replaced by some general code.

5. Refactor management of the chains to allow them to be created on-demand instead of a hardcoded list in the `MD_PARSER`, so that in the end, if it goes really well, we may get rid of almost all code for those simple marks, only `md_parse()` would test the flags, and create the respective record in step 5.

6. Then we might even potentially think about an API for registering a new simple custom mark by the application instead of modifying/extending MD4C itself everytime somebody needs a new simple span type, especially something highly specific for his use case. (But this is likely more distant future and it might require some bigger changes in the API.)

I may be wrong but I'd say at least steps 1 and 2 should be relatively easy.

(But please no crazy PR trying to do everything. One PR per step and one PR at any given time, or even in smaller steps whenever it makes sense for reviewing and testing, except perhaps steps 1 and 2 as we will need at least some code using the new generic function in order to test it and verify we go in the right direction.)

